PR #202, opened 2026-08-14
Fix authenticated hydration mismatch

### LINKED ISSUE #154: [Bug] React hydration error (#418) fires on every page load
# React hydration error (#418) fires on every page load for authenticated users

## Summary
A React hydration mismatch error fires in the browser console on every page load while logged in - confirmed live via a real browser on multiple pages. The root cause is diagnosed and the fix is concrete, not exploratory: this is the highest-impact item on the board since it affects every authenticated user, every page.

## Background
`src/hooks/useAuth.tsx`'s `AuthProvider` uses a `useState` lazy initializer (`getUserFromStorage`) that reads `localStorage` synchronously:
```ts
const [user, setUser] = useState<User | null>(getUserFromStorage);
```
`getUserFromStorage` returns `null` on the server (no `window`), but on the client's hydration pass it immediately returns the real stored user - server HTML says "logged out," the client's first render says "logged in." That mismatch is exactly what React error #418 flags. This explains why it only shows up while authenticated - an anonymous visit renders `null` consistently on both sides.

**Current behavior:** Hydration error fires on every page load for a logged-in user.
**Desired behavior:** No hydration error.

## Scope
- [ ] `getUserFromStorage`'s localStorage read moved out of the `useState` lazy initializer and into a `useEffect` (client-only, runs after hydration)
- [ ] Existing "stay logged in across a refresh" behavior still works

## Out of Scope
- Any other auth behavior changes beyond fixing this specific initialization timing

## Technical Notes
**Relevant files:**
- `src/hooks/useAuth.tsx` - `getUserFromStorage` and the `useState` call inside `AuthProvider`

**Design considerations:** A brief "logged out" flash before the effect runs is expected and fine - the hook already has a `loading` state that can gate rendering during that window if a flash is undesirable.

## Acceptance Criteria
- [ ] Confirmed via browser console (logged in) that the hydration error no longer fires
- [ ] Existing auth persistence across a page refresh still works

## Suggested Approach
1. Initialize `user` to `null` in `useState` (matching what the server renders).
2. Read from storage inside a `useEffect` and call `setUser(...)` there instead.
3. Verify via browser console on an authenticated page that the error is gone.

## Difficulty
Intermediate

## Good First Issue?
No - touches the auth provider gating every page, and the logged-out-flash tradeoff (whether to gate on `loading`) is a judgment call worth discussing.
Estimated effort: a few hours

## Contributor Checklist

Thabetahmed:
## Summary

- make the server render and initial client render use the same authentication state
- restore a valid stored session after hydration while keeping protected routes gated by loading state
- add regression coverage for authenticated hydration, anonymous sessions, and expired sessions

## Root cause

`AuthProvider` read `localStorage` from its `useState` initializer. The server rendered an anonymous state, while an authenticated browser could render a stored user during hydration, causing React error #418.

## Verification

- `npm run test:run` — 105 tests passed
- `npx tsc --noEmit`
- `npx eslint src/hooks/useAuth.tsx src/__tests__/useAuth.test.tsx`
- `npm run build`
- manually refreshed `/developer` several times while authenticated; the session persisted without a hydration error

Closes #154

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication state restoration during page loading.
  * Prevented server-rendering and client-side hydration mismatches.
  * Expired stored sessions are now cleaned up automatically.
  * Users with valid saved sessions are restored after hydration.

### DIFF
--- a/src/__tests__/useAuth.test.tsx
+++ b/src/__tests__/useAuth.test.tsx
@@ -1,0 +1,120 @@
+import { act } from 'react';
+import { hydrateRoot } from 'react-dom/client';
+import { renderToString } from 'react-dom/server';
+import { render, screen, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { AuthProvider, useAuth } from '@/hooks/useAuth';
+import type { User } from '@/types/resource';
+
+const storedUser: User = {
+  id: 42,
+  email: 'developer@example.com',
+  display_name: 'Test Developer',
+  role: 'developer',
+  created_at: '2026-08-14T00:00:00.000Z',
+};
+
+function jwtWithExp(exp: number): string {
+  const header = btoa(JSON.stringify({ alg: 'HS256' }));
+  const payload = btoa(JSON.stringify({ exp }));
+  return `${header}.${payload}.signature`;
+}
+
+function storeSession(exp: number) {
+  const token = jwtWithExp(exp);
+  localStorage.setItem('ratq_access_token', token);
+  localStorage.setItem('ratq_refresh_token', token);
+  localStorage.setItem('ratq_user', JSON.stringify(storedUser));
+}
+
+function AuthStateProbe() {
+  const { user, loading } = useAuth();
+
+  return (
+    <p data-testid="auth-state">
+      {loading ? 'loading' : user?.email ?? 'anonymous'}
+    </p>
+  );
+}
+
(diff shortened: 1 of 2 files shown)


abubakr-itqan (approved):
Verified: tsc-check clean, no new lint issues, 4/4 tests pass including a real SSR-to-hydrate regression test that would fail on the old code. Matches #154's AC exactly (useState(null) + useEffect read, loading-gated).

PR #210, opened 2026-08-15
Fix: prevent stored XSS in JsonPreview via dangerouslySetInnerHTML

usefahmed07:
json_content is publisher-supplied. The syntax highlighter built an HTML string by wrapping regex-matched tokens in <span> without escaping the token contents, then injected it with dangerouslySetInnerHTML - a JSON string value containing markup (e.g. an  tag) would execute for every visitor viewing that resource's preview.

- rewrite the highlighter to build React nodes/spans directly per token instead of an HTML string, so string content is always rendered as text and can never be parsed as markup
- also fixes a pre-existing react-hooks/error-boundaries lint error by separating JSON.parse (try/catch) from JSX construction
- add JsonPreview.test.tsx, including a regression test with an  payload asserting it renders as literal text and no  element is created

### DIFF
--- a/src/__tests__/JsonPreview.test.tsx
+++ b/src/__tests__/JsonPreview.test.tsx
@@ -1,0 +1,38 @@
+import { describe, it, expect } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { JsonPreview } from '@/modules/resources/components/preview/JsonPreview';
+
+describe('JsonPreview', () => {
+  it('renders nothing when there is no json_content', () => {
+    const { container } = render(<JsonPreview data={{}} />);
+    expect(container.innerHTML).toBe('');
+  });
+
+  it('renders formatted, syntax-highlighted JSON', () => {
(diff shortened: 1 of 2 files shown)


abubakr-itqan (approved):
Verified: real stored XSS (dangerouslySetInnerHTML fed unescaped publisher-supplied JSON string values). Fix rewrites the highlighter to build React text nodes/spans directly instead of an HTML string - content can't be parsed as markup regardless of what characters it contains. tsc-check clean, no lint issues (also fixes a pre-existing try/catch-in-JSX lint error as a bonus), 4/4 tests pass including a real  regression test. Confirmed no other dangerouslySetInnerHTML usage anywhere in the codebase - this was the only instance of this vulnerability class.

abubakr-itqan:
@usefahmed07 one small process note for next time (not blocking anything here, already merged) - branch names need one of `feature/`, `feat/`, `hotfix/`, `community/`, `chore/`, `release/`, or `dependabot/` as the prefix (CI's Branch Naming Check enforces it). `fix/` isn't in the list, which is what tripped the check on this one. Merged with an admin override since the code itself was already verified. Shukran for catching and fixing a real security issue!

PR #216, opened 2026-08-16
Fix: wrap resource detail page in withEdgeCache (#176)

### LINKED ISSUE #176: [Bug] Resource loading from the CMS is slow and the browsing experience isn't smooth
## Summary
Flagged directly in product feedback: fetching real resource data from the CMS still takes too long, and browsing/getting resources isn't smooth. This is a core-discovery-path issue - it directly affects the platform's now-top-priority experience (listing, search, sort, filter, detail page).

## Background
Some CMS-loading performance work already shipped (parallelized pagination in `cms.ts`/`payload.ts`, and a Cloudflare Workers Cache API layer for `/api/resources`/`/api/resources/[slug]` - see `DEPLOYMENT-MAP.md`'s "beta.ratq.itqan.dev/resources slow load" entry). This issue tracks re-measuring against the current experience, since the feedback suggests it's still not smooth enough - either a regression, a gap the earlier fix didn't cover (e.g. the detail page's CMS `getDetail()` call, which fetches presigned image URLs live and isn't covered by the same cache window), or a genuinely new bottleneck.

**Current behavior:** Resource loading (especially from the CMS source) feels slow / not smooth, per direct product feedback.
**Desired behavior:** Listing, searching, filtering, sorting, and viewing resource details all feel fast and responsive - this is now the platform's core value proposition, not a secondary concern.

## Scope
- [ ] Re-measure current load times for `/resources` (list) and `/resources/[slug]` (detail) against real CMS data
- [ ] Identify whether this is a cache-coverage gap (e.g. detail page image URLs), a CMS-side latency issue, or something else
- [ ] Fix the identified bottleneck(s)

## Priority
Top priority alongside the mock-data-removal issue - both are prerequisites for the core resource-discovery experience being genuinely good, per the updated product direction.

## Depends On
Nothing blocking - can start immediately, independent of the mock-data-removal issue (#138).

## Acceptance Criteria
- [ ] Documented before/after load-time measurements
- [ ] Resource browsing feels smooth in a live test, not just synthetically fast

## Difficulty
Medium-Hard - requires real investigation, not a known one-line fix.

usefahmed07:
## Summary

Partially addresses #176 by fixing the resource detail-page performance bottleneck. The list page still requires further investigation.

## Root Cause

`/resources/[slug]` was calling `resourceAggregator.get()` directly, bypassing the Workers Cache API layer (`withEdgeCache`) used by `/api/resources/[slug]`.

Since Next.js `next: { revalidate }` caching is not reliably honored on Cloudflare's Edge Runtime, detail-page requests were repeatedly hitting the CMS without benefiting from the existing edge cache.

## Fix

* Wrapped `resourceAggregator.get()` with `withEdgeCache`.
* Reused the same cache key as the API route through a synthetic `Request`, avoiding an extra self-fetch and the need for a base-URL environment variable.
* Wrapped the resource lookup with React `cache()` so `generateMetadata` and the page component share the same result within a request.
* Preserved correct handling of missing resources and added a local-development fallback when edge caching is unavailable.

### Before

**Detail page — production:**
`2.0s → 1.4s → 1.7s`

The page did not benefit from caching.

**API endpoint — production:**
`4.38s → 0.23s → 0.25s`

### After

**Detail page — local repro:**
`0.26s → 0.10s → 0.10s`

Repeated requests now benefit from edge caching.

## Notes

* Investigated `DEPLOYMENT-MAP.md` referenced in the issue background, but the file does not exist in this repository, so that specific claim could not be verified.
* This PR addresses the **detail-page** bottleneck. The `/resources` list page still needs further investigation for #176.
* Initial list-page measurement on production: `0.69s → 0.39s`.

## Testing

* `npx tsc --noEmit` — clean
* `npm run lint` — no new errors in changed files
* `npm run test:run` — **132/132 passing**

Fixes #176

### DIFF
--- a/src/__tests__/ResourceDetailPage.test.tsx
+++ b/src/__tests__/ResourceDetailPage.test.tsx
@@ -1,0 +1,150 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { Resource } from '@/types/resource';
+
+const mockGet = vi.fn();
+vi.mock('@/modules/resources/infrastructure/repositories/aggregate', () => ({
+  resourceAggregator: { get: (slug: string) => mockGet(slug) },
+}));
+
+const mockNotFound = vi.fn(() => {
+  throw new Error('NEXT_NOT_FOUND');
+});
+vi.mock('next/navigation', () => ({
+  notFound: () => mockNotFound(),
+}));
+
+vi.mock('@/app/resources/[slug]/ResourceDetailClient', () => ({
+  ResourceDetailClient: (_props: { resource: Resource }) => null,
+}));
+
+function createResource(overrides: Partial<Resource> = {}): Resource {
+  return {
+    id: 1,
+    name: 'Test Resource',
+    slug: 'test-resource',
+    type: 'api',
+    description: 'A test resource description',
+    short_description: 'Test resource summary',
+    documentation_url: null,
+    github_url: null,
+    license: 'MIT',
+    itqan_badge: false,
+    status: 'published',
+    created_at: '2024-01-01T00:00:00Z',
+    updated_at: '2024-01-01T00:00:00Z',
+    version: '1.0.0',
+    github_stats: null,
+    source: 'ratq',
+    source_url: null,
+    ...overrides,
+  } as Resource;
+}
+
+// Minimal Workers-style Cache API mock, matching what withEdgeCache expects
+// (cache.match(request) / cache.put(request, response)).
+function installMockEdgeCache() {
+  const store = new Map<string, Response>();
+  const cache = {
+    match: vi.fn(async (req: Request) => store.get(req.url)?.clone()),
+    put: vi.fn(async (req: Request, res: Response) => {
+      store.set(req.url, res);
+    }),
+  };
+  (globalThis as any).caches = { default: cache };
+  return cache;
+}
+
+function removeMockEdgeCache() {
+  delete (globalThis as any).caches;
+}
+
+beforeEach(() => {
+  vi.resetModules();
+  mockGet.mockReset();
+  mockNotFound.mockClear();
+});
+
+afterEach(() => {
+  removeMockEdgeCache();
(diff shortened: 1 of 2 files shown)


abubakr-itqan:
Took a look - the caching approach itself is sound and matches what #176's own writeup flagged as the likely gap (detail page bypassing the same edge-cache layer /api/resources/[slug] uses). tsc's clean, no lint issues in the touched file. Two things before I can merge though: the issue asks for documented before/after load-time measurements and a live-smoothness check, and the PR description is empty - can you add those? And since this introduces real caching/branching logic (cache hit vs miss, notFound path), it'd be good to have a test covering it, there isn't one for this page currently.

usefahmed07:
> Took a look - the caching approach itself is sound and matches what #176's own writeup flagged as the likely gap (detail page bypassing the same edge-cache layer /api/resources/[slug] uses). tsc's clean, no lint issues in the touched file. Two things before I can merge though: the issue asks for documented before/after load-time measurements and a live-smoothness check, and the PR description is empty - can you add those? And since this introduces real caching/branching logic (cache hit vs miss, notFound path), it'd be good to have a test covering it, there isn't one for this page currently.

Yup Yup  no problem i am on it Thanks For review

usefahmed07:
@abubakr-itqan Good Now btw

abubakr-itqan (approved):
Real before/after numbers now (detail page 2.0s/1.4s/1.7s -> 0.26s/0.10s/0.10s locally, matches the API route's own cache benefit), plus a test covering cache hit, cache miss, and the notFound path. Also noticed you tightened the cache key encoding and added a proper 404 status on the cached response along the way - nice catch. tsc clean, no lint issues, 5/5 tests pass. Appreciate you being upfront that DEPLOYMENT-MAP.md doesn't actually exist and that the list page still needs separate work - I'll reopen #176 after merge since only the detail-page half is done, so the list-page investigation doesn't get lost.

PR #224, opened 2026-08-19
feat: backend job to fetch and cache GitHub stats/commits/topics (#218)

### LINKED ISSUE #218: [Feature] Backend job to fetch and cache GitHub stats/commits/topics (extend #189 beyond per-request fetching)
## Summary
#189 wires GitHub commit history and topics into the resource detail page by fetching from `api.github.com` server-side in this Next.js app on each request (with `next: { revalidate }` caching), because there's no backend job infrastructure to do it centrally instead. This issue tracks building that backend infrastructure properly, so the per-request fetch approach can eventually be replaced/extended by it.

## Background
An older design doc (`docs/superpowers/specs/2026-04-27-versioning-github-stats-design.md`) proposed a **backend-side** approach: periodically fetch GitHub stats for resources with a `github_url`, cache the result in a `github_stats` field on the Resource model, and serve it as part of the normal API response — avoiding per-request external API calls entirely. That was never built. Verified while scoping #189:
- `github_stats` is hardcoded `null` in both frontend mapper paths (`src/shared/infrastructure/payload-resource-mapper.ts:49`, `src/modules/resources/infrastructure/repositories/cms.ts:80`)
- The Payload `Resources` collection (`payload-backend/src/collections/Resources.ts`) has no `github_stats` field at all
- `payload-backend` has no cron/scheduled-job infrastructure today (no jobs/tasks/cron directory, only CI-oriented GitHub Actions workflows)

#189 goes with per-request fetching for now because standing up this infra is real, separate scope. This issue is that scope.

## Scope
- [ ] A scheduled job (Payload job queue, or a triggered endpoint + external scheduler) that fetches stars/forks/open issues/last-commit — and, once #189 lands, commit history/topics — for every resource with a `github_url`
- [ ] Persist results on the `Resources` collection (new `github_stats`/`github_commits`/`github_topics` fields, or equivalent)
- [ ] Frontend mappers (`payload-resource-mapper.ts`, `cms.ts`) read real cached values instead of hardcoded `null`
- [ ] Uses an authenticated GitHub PAT (server-only env var), same credential-ownership pattern as `GITHUB_CLIENT_ID`/`GITHUB_CLIENT_SECRET` in `payload-backend/src/app/oauth/github/callback/route.ts`
- [ ] Reasonable refresh interval that respects GitHub's rate limits across the full resource catalog

## Out of Scope
- Changing #189's frontend components (`GithubStatsCard`, the new commits/topics component) — this is a data-source swap underneath them, not a UI change
- Real-time/on-demand refresh triggered from the frontend

## Technical Notes
- `docs/superpowers/specs/2026-04-27-versioning-github-stats-design.md` — original design intent for this
- `payload-backend/src/collections/Resources.ts` — where the cached fields would live
- `payload-backend/src/app/oauth/github/callback/route.ts` — existing pattern for owning a GitHub credential server-side

## Difficulty
Intermediate/Hard — new job infrastructure in a codebase that has none yet, plus rate-limit-aware scheduling across the full catalog.

## Depends on
#189 (frontend commit/topics UI should land first; this replaces its data source once built)

usefahmed07:
Close #218

Adds a scheduled job that fetches stars/forks/open_issues/last_commit plus recent commits and topics from GitHub for every resource with a github_url, and caches the result on the Resources collection instead of leaving github_stats hardcoded null.

- New fields on Resources: github_stats (group: stars/forks/open_issues/last_commit), github_commits, github_topics, github_stats_fetched_at
- payload-backend/src/lib/github/fetchResourceGithubData.ts: fetches repo stats + commits + topics from api.github.com using an authenticated PAT (GITHUB_STATS_TOKEN, server-only env var, same credential-ownership pattern as GITHUB_CLIENT_ID/SECRET)
- payload-backend/src/app/api/jobs/refresh-github-stats/route.ts: protected endpoint (x-job-secret header vs JOBS_SECRET) that batches through all resources with a github_url (10 at a time, 2s between batches to respect GitHub's authenticated rate limit), sorted stalest-first so incremental runs eventually cover the whole catalog
- .github/workflows/refresh-github-stats.yml: cron trigger (every 6h) that POSTs to the endpoint via curl

Frontend mappers (payload-resource-mapper.ts, cms.ts) are not touched in this PR - they already read github_stats off the Resource doc, so once this job runs in production they'll pick up real values with no code change needed there.

Requires repo secrets to be configured before the workflow can run: JOBS_SECRET, GITHUB_STATS_TOKEN (read-only PAT scoped to public repo metadata), PAYLOAD_BACKEND_URL.

**Update (migration):** added the missing Postgres migration for the four new Resources fields, per @abubakr-itqan's review comment - generated from a clean worktree off main to avoid pulling in unrelated schema drift, then merged with the latest main (announcements/notifications migrations).

## Summary by CodeRabbit

* **New Features**
  * Added automatic GitHub repository statistics, commit, and topic updates.
  * GitHub data refreshes run every six hours, with an option for manual refresh.
  * Added announcement support, including categories, descriptions, expiration, activation status, related resources, and calls to action.

* **Bug Fixes**
  * GitHub metadata is now protected from manual modification and updates safely when repository data cannot be retrieved.


abubakr-itqan (changes requested):
Shukran for tackling this - the design is solid (server-only PAT, batched with rate-limiting, restricted field access so only the job can write these). One thing's missing before this can merge though: there's no migration for the four new fields on `Resources` (`github_stats`, `github_commits`, `github_topics`, `github_stats_fetched_at`). Every other schema change in this repo has one (see the migrations folder) - without it, these columns won't actually exist in the database and the job will fail the moment it runs. Can you run `payload migrate:create` and commit the result? Once that's in, this is good to go.

abubakr-itqan:
@usefahmed07 just checking in - no rush at all, just making sure the migration ask didn't slip off your radar. Let me know if you're still working on it or if you'd like a hand with `payload migrate:create`.

usefahmed07:
Done - ran payload migrate:create and pushed the migration for the four fields. Also merged in the latest main to pick up the announcements/notifications migrations that landed in the meantime.

abubakr-itqan (changes requested):
@usefahmed07 shukran for adding the migration - the github_stats columns themselves look right (github_stats_stars/forks/open_issues/last_commit, github_commits, github_topics, github_stats_fetched_at all match Resources.ts).

One real problem though: this migration's up() also drops and re-adds 9 unrelated FK constraints (resources.owner_id, comments.author_id/resource_id, reports.resource_id/reporter_id, access_requests.resource_id/applicant_id, api_keys.resource_id/owner_id), reverting them from ON DELETE cascade back to ON DELETE set null. All 9 of those columns are required (NOT NULL) in their collection configs - that's exactly the bug #20260731_000000_cascade_required_relationships already fixed (set null on a NOT NULL column 500s the moment the referenced row gets deleted). If this merges as-is, deleting a user or resource that has any comments/reports/access-requests/api-keys would break again.

I hit this exact class of issue myself earlier today: migrate:create diffs against your local migration history, and if the migration was generated before your branch had 20260731 in it (even if you merged main in afterward), the generated file bakes in a stale diff - merging main afterward doesn't regenerate it.

Can you regenerate the migration: pull latest main first, make sure your local migrations are fully caught up (payload migrate), then run payload migrate:create again just for the github_stats fields? The new diff should only touch the resources table's new columns, nothing else. Happy to help if you get stuck.

usefahmed07:
@abubakr-itqan Good catch, but this isn't new to this PR - the disconnected prevId
(zero UUID) pattern exists in 10 of the 11 existing migrations
already (only 20260731_000000_cascade_required_relationships has a
correct prevId). This migration just follows the same
migrate:create-generated pattern the rest of the history already
has. Repairing the whole chain is a bigger, separate cleanup that
touches migrations unrelated to this PR - happy to open a follow-up
issue for it if the team wants that done.

abubakr-itqan (approved):
Confirmed - the regenerated migration only touches the seven new github_stats/github_commits/github_topics columns on resources, no FK constraint changes at all. Verified independently: tsc-check clean on both frontend and payload-backend, no new lint errors, 144/144 backend tests pass, and I tested the migration both directions locally (up creates exactly those columns, down cleanly removes them, nothing else touched). Merging.

usefahmed07:
> Confirmed - the regenerated migration only touches the seven new github_stats/github_commits/github_topics columns on resources, no FK constraint changes at all. Verified independently: tsc-check clean on both frontend and payload-backend, no new lint errors, 144/144 backend tests pass, and I tested the migration both directions locally (up creates exactly those columns, down cleanly removes them, nothing else touched). Merging.

Thanks for the thorough review and for verifying both directions of the migration!

abubakr-itqan:
@usefahmed07 good call flagging it rather than trying to fix it inline here - filed as #247. That disconnected chain is very likely why my own migrations kept regenerating wrong diffs earlier today too (announcements/notifications), so this has been on my radar as real technical debt, not just noise. No action needed from you on this PR.

usefahmed07:
> @usefahmed07 good call flagging it rather than trying to fix it inline here - filed as #247. That disconnected chain is very likely why my own migrations kept regenerating wrong diffs earlier today too (announcements/notifications), so this has been on my radar as real technical debt, not just noise. No action needed from you on this PR.

There's no problem working on it; I'm happy to be a part of this.

abubakr-itqan:
@usefahmed07 appreciate the offer, but this one's done - I dug into it myself and it turned out to be a smaller fix than the issue assumed. Wrote up the details on #247 if you're curious what was actually going on with `migrate:create`'s baseline selection (spoiler: `prevId` isn't the thing at fault). Thanks again for flagging it in the first place, that's what put it on the radar.

Issue #199, opened 2026-08-13
[Bug] Nav bar in developer view is hiding part of the sidebar


sanataff:
## What's wrong

The navbar is hiding part of the sidebar in the developer view, which make some of the sidebar content non accessible 

## Where

https://beta.ratq.itqan.dev/developer

## Steps to reproduce
Just access the url above 
## Acceptance criteria
The nav bar should not be hiding any part of the page content 

- [ ] The bug is fixed
- [ ] A regression test is added where practical

usefahmed07:
Hi @sanataff i Can work on this

abubakr-itqan:
@usefahmed07 jazak Allah khayran for picking this up and opening the PR! Assigning it to you now.

abubakr-itqan:
Closed via Notion: status changed to **Done**.

abubakr-itqan:
For the record - this contribution is now featured on the community's RATQ Contributors' Hall of Fame: https://community.itqan.dev/d/658 🎉

Issue #157, opened 2026-08-02
[Bug] Developer section top bar overlaps the global site header


abubakr-itqan:
> **Note (2026-08-03):** file paths below were updated to match the `src/` DDD restructure that landed on `beta` - the described issue itself is unaffected, only locations changed.

# Developer section top bar overlaps the global site header

## Summary
On every `/developer/*` page, the developer section's own top bar renders underneath the global site header instead of below it - both stick to the top of the viewport at the same position, so their content visually collides and becomes illegible.

## Background
The global `Header` component is `sticky top-0 z-50 h-0 overflow-visible` - a zero-height wrapper whose actual visible content floats above the page via `overflow-visible`, a pattern already used elsewhere in this codebase (e.g. `Sidebar.tsx` accounts for it with `lg:top-32`). The developer layout's own top bar is `sticky top-0 z-20` with real height and no offset - since `Header` reserves no layout space, both elements sit at `y=0`, and `Header`'s higher z-index content paints directly over the developer bar's.

**Current behavior:** Confirmed live via screenshot on `/developer/access` - both headers' text overlaps illegibly at the top of the page.
**Desired behavior:** The developer top bar sits below the global header, not underneath it.

## Scope
- [ ] The developer section's top bar no longer overlaps the global header on any `/developer/*` page
- [ ] Verified visually on at least one affected page after the fix

## Out of Scope
- Redesigning either the global header or the developer top bar's content/styling - this is a positioning fix only

## Technical Notes
**Relevant files:**
- `src/shared/ui/layout/Header.tsx` - the global header, `sticky top-0 z-50 h-0 overflow-visible`
- `src/app/developer/layout.tsx` - the developer top bar, currently `sticky top-0 z-20`
- `src/modules/developer/components/Sidebar.tsx` - already solves this same problem correctly with `lg:top-32`; use the same offset value here rather than guessing a new one

## Acceptance Criteria
- [ ] The developer top bar no longer overlaps the global header on any `/developer/*` route
- [ ] Confirmed visually (screenshot or live check)

## Suggested Approach
1. Change the developer top bar's `top-0` to the same offset `Sidebar.tsx` already uses (`top-32`) so it clears the global header's real rendered height instead of sitting at `y=0`.
2. Verify visually on `/developer/access` (or any `/developer/*` page) that the overlap is gone.

## References
None beyond the files above - the fix pattern already exists in this codebase.

## Difficulty
Beginner

## Good First Issue?
[...]

PR #225, opened 2026-08-19
fix(developer): update developer top bar offset to clear global header

rouaanouri:
Fixes the developer top bar overlap with the global site header by updating the offset from `top-0` to `top-32` in `src/app/developer/layout.tsx`.

## Changes
- Updated sticky position in `src/app/developer/layout.tsx` to `top-
image
image
32`.

## Summary by CodeRabbit

* **Style**
  * Improved the developer area’s top bar behavior by keeping it visible while scrolling on large screens.

### DIFF
--- a/src/app/developer/layout.tsx
+++ b/src/app/developer/layout.tsx
@@ -50,7 +50,7 @@ export default function DeveloperLayout({
 
       <div className="flex-grow pt-32">
         {/* Top bar */}
-        <div className="bg-[var(--bg-card)] border-b border-[var(--border-color)] px-6 py-4 flex items-center justify-between sticky top-32 z-20">
+        <div className="bg-[var(--bg-card)] border-b border-[var(--border-color)] px-6 py-4 flex items-center justify-between sticky lg:top-32 z-20">
           <div className="flex items-center gap-3">
             {/* Mobile hamburger toggle */}
             <button


abubakr-itqan (changes requested):
Thanks for picking this up! Two things worth flagging before this can merge.

First, a real regression: this breaks `DeveloperLayout.test.tsx`'s existing test ("offsets the sticky top bar below the fixed root header (regression for #199)"), which explicitly asserts `top-32` unconditionally and guards against exactly this kind of change. Scoping it to `lg:top-32` means on mobile/tablet widths the top bar loses its offset entirely and the original overlap comes back - the same failure mode #199 was written to prevent.

Second, and this might make the fix unnecessary altogether: I checked `main`, and the developer top bar already has `sticky top-32 z-20` (not `top-0` as #157 describes) - it looks like #157 might be a stale duplicate of #199, which already fixed this exact overlap and shipped with that regression test. Could you check live on `/developer` whether you're still actually seeing an overlap on any screen size? If not, this issue is probably just outdated and can be closed instead of needing a PR at all - no fault of yours, the issue's description just didn't get updated when #199 landed.

rouaanouri:
Thank you for the detailed feedback!

I checked main and verified that top-32 is indeed already applied unconditionally, so the overlap issue (#157) is already fixed by #199 as you mentioned.

Since this issue is obsolete, I will close this PR. Thanks again for your guidance!

Issue #155, opened 2026-08-02
[Enhancement] Mount Announcements and Trending Resources on the homepage


abubakr-itqan:
> **Note (2026-08-03):** file paths below were updated to match the `src/` DDD restructure that landed on `beta` - the described issue itself is unaffected, only locations changed.

# Mount Announcements and Trending Resources on the homepage

## Summary
`AnnouncementsCarousel` and `TrendingResources` are both built and neither is rendered anywhere in the app. Announcements is fully working and just needs to be wired in. Trending Resources has a real crash bug (tracked separately) that must be fixed first - mounting it before that fix ships a broken homepage.

**Blocked on:** the `TrendingResources` crash fix (companion issue) must land before this one, for the Trending half specifically. The Announcements half has no such dependency and can proceed independently.

## Background
`AnnouncementsCarousel.tsx` and `TrendingResources.tsx` both exist, fully built (carousel auto-rotation/ARIA/keyboard-nav for the former, a period toggle for the latter). `AnnouncementsCarousel` and its data hook are fully tested and green. `TrendingResources` currently crashes on real data (see the companion crash issue) - fix that first.

Neither component is imported anywhere under `src/app`. The repo's git history for `src/app/page.tsx` shows a single commit on this branch, so this reads as "built but never wired into this page" rather than a regression from an earlier working state - the homepage was written fresh during the beta rebuild and these two pieces weren't included.

**Current behavior:** Neither section appears on the homepage.
**Desired behavior:** Both appear on the homepage, in a position a maintainer signs off on.

## Scope
- [ ] `AnnouncementsCarousel` mounted on the homepage
- [ ] `TrendingResources` mounted on the homepage, **after** its crash fix has landed
- [ ] Placement confirmed with a maintainer before implementing - there's no existing "Featured Resources" section on the current homepage to anchor against, so this needs an explicit decision, not just copying an old design spec's position

## Out of Scope
- Redesigning either component
- Building a real (non-mock) backend for announcements or trending - separate, larger work
- Any other homepage content changes beyond adding these two sections

## Technical Notes
**Relevant files:**
- `src/app/page.tsx` - current homepage, needs both components added
- `src/modules/resources/components/AnnouncementsCarousel.tsx`, `src/modules/resources/components/TrendingResources.tsx`
- `src/hooks/useAnnouncements.ts`, `src/hooks/useTrendingResources.ts`

## Acceptance Criteria
- [ ] Homepage renders the announcements carousel when active announcements exist
[...]

PR #236, opened 2026-08-19
Add announcements section

boss-moh:
Replace the old placeholder announcements with announcements carousel 

## Summary by CodeRabbit

- **New Features**
  - Added an announcements carousel to present announcements in a more interactive format.
- **Improvements**
  - Replaced the previous static announcement banner with the new carousel experience.

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -7,6 +7,7 @@ import { useGSAP } from "@gsap/react";
 import { useLanguage } from "@/shared/ui/i18n";
 import { useResources } from "@/hooks/useResources";
 import { ResourceCard } from "@/modules/resources/components/ResourceCard";
+import AnnouncementsCarousel from "@/modules/resources/components/AnnouncementsCarousel";
 
 gsap.registerPlugin(useGSAP);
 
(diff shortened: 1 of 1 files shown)


boss-moh:
@abubakr-itqan 
I have finished the announcements  
therefor i start work on  Trending but the <TrendingResources/> components doesn't display any data 
(I solve the problem of rendering but i don't have data to render)

as what I understand i should replace the old trending UI with this component

boss-moh:
> @abubakr-itqan I have finished the announcements therefor i start work on Trending but the components doesn't display any data (I solve the problem of rendering but i don't have data to render)
> 
> as what I understand i should replace the old trending UI with this component

I found the problem old trending's download count is zero and there are an conidiation that filter non download libs (zero )

There a few questions: 
what we should display when 
- there are no trending items back from API
- if there are happen a error in the process of  fetching data

abubakr-itqan (approved):
Verified: tsc-check clean, no new lint errors (checked against main's baseline - same 20 pre-existing problems, none new), and the existing AnnouncementsCarousel/useAnnouncements test suites all pass (10/10). Diff is exactly the swap #155 calls for - the static placeholder banner replaced with the real carousel component, right where it was. Merging.

PR #238, opened 2026-08-19
Add TrendingResources 

boss-moh:
update the home page add Trending Resources and replace the old code 
I have did : 
- add a fallback state instead of remove the component and doesn't display it in the page 
- add a arabic and english translation 

note: the UI will display the fallback beacuse there are no items that fit the filter or the conidiation (at file 
src\modules\resources\domain\services\trending-ranking.ts) 
`
    items.filter((r) => (isAllTime ? r.total_downloads > 0 : r.downloads > 0))
`

    there are no items with download you should either edit the data or the conidiation
    
The UI when there are 
- data 
image

- no data
image

## Summary by CodeRabbit

* **New Features**
  * Added announcements and trending resources to the home page.
  * Expanded resource descriptions to display up to three lines.

* **Bug Fixes**
  * Improved handling when no trending resources are available.

* **Tests**
  * Updated resource card coverage for three-line description truncation.

### DIFF
--- a/src/__tests__/ResourceCard.test.tsx
+++ b/src/__tests__/ResourceCard.test.tsx
@@ -112,13 +112,13 @@ describe('ResourceCard', () => {
     expect(screen.queryByText('Demo')).not.toBeInTheDocument();
   });
 
-  it('clamps short description to 2 lines', () => {
+  it('clamps short description to 3 lines', () => {
     renderWithProvider(
       <ResourceCard resource={createResource({
-        short_description: 'This is a very long description that should be clamped to two lines maximum regardless of how much text is provided here to ensure the card stays compact.',
+        short_description: 'This is a very long description that should be clamped to three lines maximum regardless of how much text is provided here to ensure the card stays compact.',
       })} />
     );
     const descriptionEl = screen.getByText(/This is a very long description/i);
-    expect(descriptionEl.closest('p')).toHaveClass('line-clamp-2');
+    expect(descriptionEl.closest('p')).toHaveClass('line-clamp-3');
   });
 });
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -7,6 +7,8 @@ import { useGSAP } from "@gsap/react";
 import { useLanguage } from "@/shared/ui/i18n";
 import { useResources } from "@/hooks/useResources";
 import { ResourceCard } from "@/modules/resources/components/ResourceCard";
+import AnnouncementsCarousel from "@/modules/resources/components/AnnouncementsCarousel";
+import TrendingResources from "@/modules/resources/components/TrendingResources";
 
 gsap.registerPlugin(useGSAP);
 
@@ -71,9 +73,8 @@ export default function HomePage() {
   const { t, direction } = useLanguage();
   const home = t.home;
 
-  const { data: trendingData, isLoading: trendingLoading } = useResources({ sort: "downloads", page_size: 4 });
   const { data: featuredData, isLoading: featuredLoading } = useResources({ sort: "newest", page_size: 4 });
-  const trendingResources = trendingData?.results ?? [];
+
   const featuredResources = featuredData?.results ?? [];
   const stats: Stat[] = [
     { value: direction === "rtl" ? "12.4 ألف" : "12.4k", label: home.stats.downloads },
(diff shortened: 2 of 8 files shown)


abubakr-itqan (changes requested):
Jazak Allah khayran for taking this on, especially finding the root cause of the empty-trending issue - that `downloads > 0`/`total_downloads > 0` filter in trending-ranking.ts really is why nothing shows for real (non-mock) resources, since there's no download tracking on the CMS side yet. And nice catch fixing the hooks-order bug too - useMemo was being called conditionally before your change (after an early `return null`), which is exactly the kind of thing that can crash React at runtime. That's a real fix, not just cleanup.

Two things need attention before this can merge though:

1. Two existing tests now fail: `ResourceCard.test.tsx`'s "clamps short description to 2 lines" (you changed `line-clamp-2` to `line-clamp-3` on ResourceCard - looks unrelated to trending, was that intentional? if so the test needs updating, but I'd rather understand why first) and `TrendingResources.test.tsx`'s "returns null when no resources and not loading" (this one's expected, since you deliberately changed the behavior from hiding the section to showing a fallback message - but that's a real design change from how the rest of the homepage handles empty sections, e.g. Featured just hides itself when there's nothing to show).

2. On the design question you asked on #236 - showing a "no trending resources" message vs. hiding the section: given the download-count issue means this will show empty basically always in production right now (real CMS resources don't have download tracking yet), I want to loop in a maintainer decision on this specific point before it merges, since it changes what visitors see on the live homepage. Hold off on this one for now - I'll get back to you on the fallback-vs-hide call.

For now: can you revert the ResourceCard.tsx line-clamp change (or explain why it's needed) and I'll follow up on the fallback design question separately.

boss-moh:
Ooh, sorry.
It slipped my mind, to run the testing code.

1. The reason behind changing line-clamp from 2 to 3 it is UI show a part of the line and also hidden a part like the following

- line-clamp-2 

image

- so I have tried to fix it line-clamp-3

image


Therefore Tests supposed to be updated


 

2.  in the part (homepage handles empty sections, e.g. Featured just hides itself when there's nothing to show)
so the behavior to hidden sections when there is no data but what if all sections hidden arbitrarily for some reason.
Therefore there is no thing to show for the client and he will thought the website are broken

abubakr-itqan:
@boss-moh good find on the line-clamp issue - checked your screenshots and you're right, line-clamp-2 is genuinely rendering a broken partial third line there, not just truncating cleanly. That's a real bug, not scope creep - line-clamp-3 is the correct fix. Go ahead and update the ResourceCard test to expect line-clamp-3 instead of line-clamp-2 and that one's resolved.

On the empty-state question - I hear you on the "whole homepage looks broken" concern, but I checked with a maintainer and we're going to keep it consistent with how Featured and the rest of the homepage already behave: hide the section when there's nothing to show, rather than a fallback message. Realistically it's very unlikely Trending, Featured, and Announcements would all be empty at the same time, and mixing "sometimes shows a message, sometimes just disappears" across sections is more confusing than reassuring. Can you swap the fallback back to the original `return null` when `resources.length === 0`? Once both of these are in, this should be good to merge.

boss-moh:
Sound great,
I'll do two thing 
- hidden the section in no data case 
- update the testing part

boss-moh:
@abubakr-itqan 
Now , everything run well

abubakr-itqan:
@boss-moh both fixes look right - tested and confirmed: line-clamp-3 test passes, and Trending now correctly hides when there's no data (and the hooks-order fix is still intact). Nice work.

Two things before this can merge though:

1. This branch now conflicts with main - #236 (a different PR) already merged the exact same Announcements carousel wiring into page.tsx while this one was open. You'll need to rebase/merge main into your branch and resolve that.
2. I noticed you renamed `TrendingPeriod` to `PeriodType` across several files (the hook, the ranking service, the API layer, the types file). What's the reasoning for that? It's not something I asked for and it's unrelated to the trending fix - if there's no specific reason, let's leave that rename out and keep this PR scoped to the trending work.

boss-moh:
About the point number 2 : just to make the things more clear by using clear name, the purpose may some parts will using the same PeriodType 
I'll reused the old name 
Sorry for that

PR #245, opened 2026-08-21
fix(auth): handle expired and invalid sessions during API requests - PR against issue #165

### LINKED ISSUE #165: [Bug] Expired session tokens fail silently instead of prompting re-login
> **Note (2026-08-03):** file paths below were updated to match the `src/` DDD restructure that landed on `beta` - the described issue itself is unaffected, only locations changed.

# Expired session tokens fail silently instead of prompting re-login

## Summary
There is no session-expiry handling anywhere in the app. Once a user's Payload JWT expires, the UI still shows them as logged in - `useAuth` only checks whether a cached user object exists in `localStorage`, never whether the token is still valid. Every subsequent write (create/update/delete) then fails with Payload's generic "You are not allowed to perform this action" message, with no indication to the user that their session expired or that logging back in would fix it.

Reproduced live: attempting to save a new resource from `/dashboard/resources` returned this exact error while the UI showed the user as fully logged in.

## Background
`src/hooks/useAuth.tsx`'s `getUserFromStorage` only reads a cached `User` object from `localStorage` - it never checks the token's own validity or expiry. `src/shared/infrastructure/token-storage.ts`'s `getAccessToken()` just returns whatever string is in `localStorage`, with no expiry check either. Payload issues short-lived JWTs (session `expiresAt` values seen elsewhere in this project are on the order of a couple of hours) with no refresh-token mechanism in this codebase (`api-client.ts`'s own comment: *"Payload issues a single JWT (no separate refresh token)"*). **Update:** a proactive check now exists - `src/modules/auth/domain/auth-token.ts`'s `AuthToken.isExpired()`, called from `useAuth.tsx`, clears a stale session on load. The gap below (redirect-to-login on a 401 *during* a write, distinguished from a real permissions denial) is still open.

Collection access checks like `Resources.ts`'s `create: ({ req }) => Boolean(req.user)` only fail when the request's JWT doesn't resolve to a user - which is exactly what happens once the token has expired. The resulting error is indistinguishable, from the user's perspective, from a real permissions problem.

**Current behavior:** An expired session shows no warning; write operations fail with a generic, misleading error message.
**Desired behavior:** The app detects an expired/invalid session and prompts the user to log back in, rather than surfacing a permissions-sounding error for what is actually a session problem.

## Scope
- [ ] API calls that fail with an auth-related error (401/403 specifically tied to an invalid/expired token, not a genuine permissions denial) trigger a clear "your session expired, please log in again" flow
- [ ] The user is redirected to `/login` (or shown a clear prompt) rather than silently failing whatever action they were attempting

## Out of Scope
- Building a real refresh-token mechanism (Payload's local auth strategy in this project issues one JWT, no refresh token - that's a bigger backend change, not required to fix the UX gap)
- Any change to Payload's own token lifetime/expiry configuration

## Technical Notes
**Relevant files:**
- `src/hooks/useAuth.tsx` - `getUserFromStorage`, has no expiry awareness
- `src/shared/infrastructure/token-storage.ts` - `getAccessToken`
- `src/shared/infrastructure/payload-error.ts` - `payloadErrorMessage` (the generic error-message helper every write function already calls on a failed response - this is the natural place to detect an auth-specific failure and react to it)

**Design considerations:** Distinguishing "your token expired" from "you genuinely don't have permission" matters for the message shown to the user - both currently produce the same generic Payload error text. Payload's own response body/status code on an auth failure vs. a genuine access-denied failure may not differ enough to tell them apart cleanly; worth checking Payload's actual response shape for an expired-JWT request specifically (its JWT verification failure vs. its access-control-function-returned-false failure may carry different status codes even if the surfaced message text looks similar).

## Acceptance Criteria
- [ ] A write attempt with an expired token results in a clear message + redirect to login, not a bare permissions-style error
- [ ] A write attempt that is genuinely disallowed (real permissions denial) still shows an appropriate message, distinct from the session-expiry case
- [ ] Manually verified by letting a session expire (or clearing the stored token) and attempting a write

## Suggested Approach
1. Investigate Payload's actual HTTP response (status code and body) for a request with an expired/invalid JWT vs. one that's genuinely denied by an access-control function, to find a reliable way to distinguish the two server-side.
2. Add a shared response-handling helper (or extend `payloadErrorMessage`) that checks for the auth-expiry case specifically and triggers a redirect-to-login / clear "session expired" message instead of the generic error text.
3. Manually test by clearing/expiring a token and confirming the new behavior.

## Difficulty
Intermediate

## Good First Issue?
No - requires investigating Payload's actual error responses to reliably distinguish two failure modes, not just a UI tweak.
Estimated effort: half a day

## Contributor Checklist

Ramahadam:
## Description

Closes #165

## Summary

This PR adds session-expiry handling for authenticated API requests.

Previously, an expired or invalid JWT could cause a protected API request to fail without consistently clearing the stale authentication state or guiding the user back to login.

The implementation now detects authentication-related API failures, verifies whether the current session is still valid, and triggers the session-expired flow only when the session is actually invalid.

This also preserves the distinction between:

* an expired or invalid session
* a valid authenticated session that does not have permission to perform an action

## Changes

* Add shared session-expiry event handling and subscriptions.
* Detect authentication-related `401` / `403` API responses.
* Probe the current session before treating an authorization failure as session expiry.
* Clear stale authentication state when the session is confirmed invalid.
* Redirect users with expired or invalid sessions to `/login`.
* Display a localized session-expired message after redirection.
* Pass authentication context through affected API infrastructure modules.
* Add tests for:

  * `payloadErrorMessage`
  * `useAuth`
  * `LoginForm`

## Session-expiry flow

```text
Protected API request
        ↓
401 / 403 response
        ↓
Check current session
        ↓
Session invalid
        ↓
Emit session-expiry event
        ↓
Clear authentication state
        ↓
Redirect to /login
        ↓
Display session-expired message
```

A genuine permission denial follows a different path:

```text
Protected API request
        ↓
403 Forbidden
        ↓
Check current session
        ↓
Session still valid
        ↓
Do not expire session
        ↓
Keep user logged in
        ↓
Display permission error
```

## Manual Testing

The following scenarios were verified manually.

### Expired JWT

Generated a correctly signed JWT with an expiration timestamp in the past and used it for an authenticated API request.

Result:

* API request rejected
* session validation failed
* authentication state cleared
* user redirected to `/login`
* session-expired message displayed

✅ Passed

### Invalid JWT

Modified the stored JWT so that the token was no longer valid and triggered a protected API request.

Result:

* invalid session detected
* authentication state cleared
* user redirected to `/login`
* session-expired message displayed

✅ Passed

### Valid session with genuine permission denial

Logged in using a valid Publisher account and attempted to update a resource owned by another user.

Observed:

```text
PATCH protected resource → 403
GET /api/users/me        → 200
```

Result:

* permission error displayed
* Publisher remained authenticated
* no session-expiry event triggered
* no redirect to `/login`

✅ Passed

This verifies that genuine authorization failures are not incorrectly treated as expired sessions.

## Automated Testing

After rebasing onto the latest `beta`:

```text
Test Files  31 passed (31)
Tests       180 passed (180)
```

TypeScript validation also passes:

```bash
npx tsc --noEmit
```

## Commits

```text
91ab938 fix(auth): detect expired sessions from API responses
22f2f3c fix(auth): handle session expiry and redirect to login
```

## Related UX Finding

While testing the new session-expired message, a separate authentication UI issue was discovered.

Authentication errors can persist when navigating between the login and registration pages. The behavior is not specific to session expiry; it can also be reproduced using a normal invalid-credentials error such as:

`The email or password provided is incorrect.`

Because this is a broader authentication error-state lifecycle issue and can be reproduced independently of session-expiry handling, it is intentionally **not included in this PR**.

It should be tracked separately so that #165 remains focused on expired/invalid session detection and handling.

## Checklist

* [x] Expired JWT redirects to login
* [x] Invalid JWT redirects to login
* [x] Session-expired message is displayed
* [x] Authentication state is cleared on confirmed session expiry
* [x] Genuine `403` permission errors do not log out valid users
* [x] Permission errors remain visible to the authenticated user
* [x] Unit tests added/updated
* [x] Full test suite passes
* [x] TypeScript check passes
* [x] Branch rebased onto latest `beta`

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection of expired or invalid sessions.
  * Users are signed out and redirected to the login page when their session expires.
  * Added localized session-expiration messages in English and Arabic.
  * Improved error handling for authenticated requests to preserve accurate session and permission errors.

* **Tests**
  * Added coverage for localized messages, session expiry, redirects, token validation, and error handling.

### DIFF
--- a/src/__tests__/LoginForm.test.tsx
+++ b/src/__tests__/LoginForm.test.tsx
@@ -1,0 +1,38 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { LoginForm } from '@/modules/auth/components/LoginForm';
+import { SESSION_EXPIRED_REASON } from '@/shared/infrastructure/session-expiry';
+import { LanguageProvider } from '@/shared/ui/i18n';
+
+vi.mock('next/navigation', () => ({
+  useRouter: () => ({ push: vi.fn(), refresh: vi.fn() }),
+}));
+
+vi.mock('@/hooks/useAuth', () => ({
+  useAuth: () => ({
+    login: vi.fn(),
+    loading: false,
+    error: SESSION_EXPIRED_REASON,
+  }),
+}));
+
+describe('LoginForm session expiry message', () => {
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
+  it.each([
+    ['en', 'Your session has expired. Please log in again.'],
+    ['ar', 'انتهت صلاحية جلستك. يُرجى تسجيل الدخول مرة أخرى.'],
+  ] as const)('shows the session-expired reason in %s', async (locale, message) => {
+    localStorage.setItem('ratq_locale', locale);
+
+    render(
+      <LanguageProvider>
+        <LoginForm />
+      </LanguageProvider>
+    );
+
+    expect(await screen.findByText(message)).toBeInTheDocument();
+  });
+});
--- a/src/__tests__/payloadError.test.ts
+++ b/src/__tests__/payloadError.test.ts
@@ -1,0 +1,234 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { payloadErrorMessage } from '@/shared/infrastructure/payload-error';
+import { PAYLOAD_API_BASE } from '@/shared/infrastructure/payload-config';
+import { subscribeToSessionExpiry } from '@/shared/infrastructure/session-expiry';
+
+const permissionMessage = 'You are not allowed to perform this action.';
+
+function jwtWithExp(exp: number): string {
+  const header = btoa(JSON.stringify({ alg: 'HS256' }));
+  const payload = btoa(JSON.stringify({ exp }));
+  return `${header}.${payload}.signature`;
+}
+
+function errorResponse(status: number, message: string): Response {
+  return new Response(JSON.stringify({ errors: [{ message }] }), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
+describe('payloadErrorMessage session validation', () => {
+  const onSessionExpired = vi.fn();
+  let unsubscribe: (() => void) | undefined;
+
+  beforeEach(() => {
+    localStorage.clear();
+    onSessionExpired.mockClear();
+    unsubscribe = subscribeToSessionExpiry(onSessionExpired);
+  });
+
+  afterEach(() => {
+    unsubscribe?.();
+    vi.unstubAllGlobals();
+    vi.useRealTimers();
+  });
+
+  it('signals session expiry when Payload rejects an expired stored token', async () => {
+    const token = jwtWithExp(Math.floor(Date.now() / 1000) - 60);
+    localStorage.setItem('ratq_access_token', token);
+    const fetchMock = vi.fn().mockResolvedValue(
+      new Response(JSON.stringify({ user: null }), {
+        status: 200,
+        headers: { 'Content-Type': 'application/json' },
+      })
+    );
+    vi.stubGlobal('fetch', fetchMock);
+
+    await payloadErrorMessage(errorResponse(403, permissionMessage), 'Failed to publish resource', {
+      authenticated: true,
+    });
+
+    expect(fetchMock).toHaveBeenCalledWith(`${PAYLOAD_API_BASE}/users/me`, {
+      cache: 'no-store',
+      headers: { Authorization: `JWT ${token}` },
+      signal: expect.any(AbortSignal),
+    });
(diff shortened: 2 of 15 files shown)


abubakr-itqan (approved):
Really thorough work, and the distinction is exactly right - probing /users/me before treating a 401/403 as session expiry is the correct way to avoid logging out a genuinely-authenticated user who just hit a permissions wall. Verified independently: tsc-check clean, no new lint errors, and the full suite passes (180/180, matches what you reported). The timeout/abort handling and the race-condition test (session token replaced mid-probe) are exactly the kind of edge cases that are easy to skip and easy to regret skipping. Appreciate flagging the login/register error-persistence issue separately instead of scope-creeping it in here - that's #246, already assigned. Merging.

Issue #257, opened 2026-08-23
[Chore] Frontend tsc has no CI check - type errors on main go undetected


abubakr-itqan:
Found while reviewing PR #256: running `tsc --noEmit` on `main` (before that PR's changes) already fails with pre-existing errors, unrelated to any change in flight:

```
src/__tests__/ForgotPasswordPage.test.tsx(44,42): error TS2352: Conversion of type '{ push: Mock<Procedure>; }' to type 'AppRouterInstance' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
src/__tests__/LoginPage.test.tsx(56,40): error TS2345: Argument of type '{ user: any; loading: false; error: null; login: Mock<Procedure>; loginWithToken: Mock<Procedure>; register: Mock<Procedure>; logout: Mock<...>; }' is not assignable to parameter of type 'AuthContextType'.
src/__tests__/LoginPage.test.tsx(73,40): error TS2345: Argument of type '{ user: null; loading: false; error: null; login: Mock<Procedure>; loginWithToken: Mock<Procedure>; register: Mock<Procedure>; logout: Mock<...>; }' is not assignable to parameter of type 'AuthContextType'.
src/__tests__/LoginPage.test.tsx(90,40): error TS2345: Argument of type '{ user: null; loading: true; error: null; login: Mock<Procedure>; loginWithToken: Mock<Procedure>; register: Mock<Procedure>; logout: Mock<...>; }' is not assignable to parameter of type 'AuthContextType'.
```

These test files mock `useRouter()` and `useAuth()` with object literals that don't satisfy the real `AppRouterInstance`/`AuthContextType` shapes (likely left over from PR #253's auth flows work). They slipped in undetected because there's no frontend typecheck step in CI - only `eslint` and `vitest` run, and vitest doesn't type-check test files it executes.

This is the same class of gap as #250, which added a payload-backend typecheck to CI after a type error slipped through there too - the frontend side has the identical hole.

## Acceptance criteria
- [ ] Fix the four existing type errors in `ForgotPasswordPage.test.tsx` and `LoginPage.test.tsx` (properly typed mocks, not `as any`/double-casts)
- [ ] Add a `tsc --noEmit` step to the frontend CI workflow so a type error in app or test code fails the PR, mirroring what #250 did for payload-backend

usefahmed07:
Hi @abubakr-itqan Can you assign it to me?

[...]

PR #249, opened 2026-08-22
Feat: add image support to resources (#226)

usefahmed07:
Closes #226

## Summary
Adds image/screenshot support for `payload`-sourced resources — the `resources` table, the self-serve publish/edit form, and both render surfaces (catalog card + detail page) had no way to carry a visual for a resource. This PR wires an optional `image` field end-to-end: backend collection field → migration → upload form → catalog card → detail page.

CMS-sourced resources are untouched — they already have images via `cms.itqan.dev/gallery/asset/{id}`, which is a separate, existing mechanism and out of scope here.

## Changes

**Backend**
- `Resources.ts`: added an optional `image` field (`upload` type, `relationTo: 'media'`) — not required, so existing resources are unaffected
- New migration `20260822_232715_add_resources_image`: adds nullable `image_id` FK to `media(id)` with `ON DELETE set null` (so a deleted media doc doesn't break the resource), plus an index. Reversible via `down`.
- `payload-types.ts` regenerated via `generate:types`

**Frontend**
- `types/resource.ts`: added optional `image_url` to `Resource`
- `payload-resource-mapper.ts`: maps Payload's populated media relationship (or bare id/null) to `image_url`
- `resources-api.ts`: added `uploadMedia()` to POST a file to Payload's `/media` endpoint; `CreateResourceInput` now accepts `image?: number | null`
- `ResourceForm.tsx`: file input + live preview; uploads the image (if changed) before create/update, disables submit while uploading, surfaces upload errors separately from the main form error
- `dashboard/resources/page.tsx`: threads `image`/`image_url` through `handleCreate`/`handleEdit` and edit-initial-values
- `ResourceCard.tsx`: renders `image_url` as a card thumbnail when present
- `ResourceDetailClient.tsx`: renders `image_url` as a hero image above the header when present

In every case the image is purely additive — when `image_url` is absent, the existing text-only layout renders exactly as before.

## Testing
- `npx tsc --noEmit` — clean
- `npm test` — 180/180 passing
- `eslint` on all changed files — no new errors. One pre-existing `react-hooks/set-state-in-effect` error surfaces on `dashboard/resources/page.tsx` line 29 (the `useEffect` calling `setIsClient`/`router.push`) but it's untouched by this diff (`git diff` shows only 3 line changes in that file, none near line 29) — leaving it out of scope for this PR.
[...]

PR #259, opened 2026-08-23
fix(register): enforce 8-64 character passwords on register

### LINKED ISSUE #255: [Bug] Registration has no password length validation
## Summary
Registration has no password length validation anywhere - not client-side, not server-side. The new reset-password flow (#253) added an 8-64 character client-side check, but a user can still register with a 1-character (or several-thousand-character) password, since neither `RegisterForm.tsx` nor Payload's `Users` collection (`auth: true`, no length constraints configured) enforce anything.

## Background
Found by @ekafe while implementing #253's reset-password validation - see [this comment](https://github.com/Itqan-community/RATQ/pull/253#issuecomment-5385178714). Confirmed independently: Payload 3.86's core auth has no default `minLength`/`maxLength` on password, and this repo's `Users.ts` doesn't set one either, so `auth: true` leaves it completely unconstrained.

## Scope
- [ ] Add the same 8-64 character validation to `RegisterForm.tsx` that `ResetPasswordForm.tsx` already has, so both entry points agree
- [ ] Decide whether that constraint should also be enforced server-side (e.g. a custom `validate` on a `beforeValidate` hook on `Users`, since Payload itself won't do this for you) so the API can't be used to bypass the frontend check entirely

## Out of Scope
- Broader password strength rules (complexity, breach-list checks, etc.) - just closing the length gap

## Acceptance Criteria
- [ ] Registering with a password shorter than 8 or longer than 64 characters is rejected with a clear error, both via the UI and directly against the API
- [ ] Existing accounts are unaffected (this only changes validation on new password creation)

## Difficulty
Easy - the client-side check already exists as a pattern to copy from `ResetPasswordForm.tsx`; the server-side piece needs a small `Users.ts` hook.

ekafe:
## Summary
- Register uses the same 8–64 password length check as reset (`validatePassword`).
- Users `beforeValidate` enforces the same range on the API so the check can’t be skipped.
- Updates with no `password` field are skipped so existing accounts / profile edits stay valid.

Closes https://github.com/Itqan-community/RATQ/issues/255

image

## Test plan
- [ ] Register with password length < 8 or > 64 → clear error, no request
- [ ] Register with 8–64 characters → succeeds (or Payload error if email verify blocks login)
- [ ] `POST /api/users` with a too-short password → 422
- [ ] Update a user without sending `password` → still works

### DIFF
--- a/payload-backend/src/collections/Users.ts
+++ b/payload-backend/src/collections/Users.ts
@@ -1,4 +1,5 @@
 import type { CollectionConfig, FieldAccess } from 'payload'
+import { passwordValidation } from './hooks/validate-password'
 
 // Mirrors the itqan_badge pattern in Resources.ts - role is a privilege
 // escalation vector, not self-declared metadata, so only an existing admin
@@ -12,6 +13,9 @@ export const Users: CollectionConfig = {
     useAsTitle: 'email',
   },
   auth: true,
+  hooks: {
+    beforeValidate: [passwordValidation],
+  },
   access: {
     // Public self-registration (no default access config would otherwise
     // require an authenticated user even to create an account).
--- a/payload-backend/src/collections/hooks/validate-password.test.ts
+++ b/payload-backend/src/collections/hooks/validate-password.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from 'vitest'
+import { APIError } from 'payload'
+import { passwordValidation } from './validate-password'
+
+async function runHook(data: Record<string, unknown> | undefined) {
+  return passwordValidation({ data } as Parameters<typeof passwordValidation>[0])
+}
+
+describe('passwordValidation', () => {
+  it('returns without throwing when no data is provided', async () => {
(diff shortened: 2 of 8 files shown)


abubakr-itqan (approved):
Reviewed properly - checked out the branch, ran payload-backend's own tsc (learned that lesson from #249) plus its own vitest suite (160/160 pass, including the new validate-password hook tests), and the full frontend suite (38 files, 220 tests pass). Lint clean on the touched files - the frontend errors I'm tracking separately (#257) are pre-existing and untouched by this diff.

Matches #255's AC on both sides: client-side check in RegisterForm mirrors ResetPasswordForm's existing pattern, server-side beforeValidate hook on Users closes the API bypass, and the final revision correctly treats a missing password as a no-op (profile updates without a password change still work) rather than rejecting it outright - good catch on your own end during the CR pass. The two failing checks (detect/detect, notify/pr-opened) are the known bot-permission 403s, not a real problem. Merging.

Issue #250, opened 2026-08-23
[CI] payload-backend has no typecheck/PR-time checks — invalid TS can merge undetected


usefahmed07:
## Summary
`payload-backend/` currently has no CI check that runs on pull requests. `deploy-backend.yml` only triggers on `push` to `main`/`beta`, and even then it only runs `pnpm test` (vitest) - no `tsc --noEmit`. `frontend-tests.yml` explicitly excludes `payload-backend/**` via `paths-ignore`.

## Background
Found via CodeRabbit review on #249: a collection config used an invalid `upload.limits` key (not a real option on Payload v3.86's collection-level upload config). `npx tsc --noEmit` from the frontend root didn't catch it since it doesn't check the backend's own `tsconfig.json`, and CI didn't catch it either since no PR-time job compiles `payload-backend/`. The invalid code would have merged and only failed at runtime/build.

## Scope
- Add a `typecheck` script to `payload-backend/package.json` (`tsc --noEmit`)
- Add a new `backend-tests.yml` workflow (or extend `deploy-backend.yml`'s `test` job) that runs on `pull_request` for `payload-backend/**` paths, running `pnpm install --frozen-lockfile` + `pnpm run typecheck` + `pnpm test`

## Acceptance Criteria
- A PR touching `payload-backend/` with a TypeScript error fails CI
- Existing `deploy-backend.yml` deploy behavior is unaffected

abubakr-itqan:
Closed via Notion: status changed to **Done**.

abubakr-itqan:
For the record - this contribution is now featured on the community's RATQ Contributors' Hall of Fame: https://community.itqan.dev/d/658 🎉

PR #260, opened 2026-08-24
Fix(tests): properly type router/auth mocks; ci: add frontend typeche…

### LINKED ISSUE #257: [Chore] Frontend tsc has no CI check - type errors on main go undetected

usefahmed07:
## Summary

Fixes the pre-existing `tsc --noEmit` failures on `main` and closes the CI gap that let them land undetected — the frontend counterpart to #250.

Closes #257

## Changes

**1. Fixed the 4 pre-existing type errors** in `ForgotPasswordPage.test.tsx` and `LoginPage.test.tsx`:
- Both files were mocking `useRouter()` with a partial object literal cast via `as any` / `as ReturnType<typeof useRouter>`, and `LoginPage.test.tsx` was mocking `useAuth()` the same way — these casts bypassed type-checking instead of satisfying it, so the mocks silently drifted out of sync with `AuthContextType` after auth flows work landed in #253 (which added `forgotPassword`, `resetPassword`, `verifyEmail`).
- Replaced both with typed factory helpers under `src/__tests__/test-utils/`:
  - `createMockRouter(overrides?)` — returns a full `ReturnType<typeof useRouter>` object with every method (`push`, `replace`, `refresh`, `back`, `forward`, `prefetch`) as `vi.fn()`, so overrides are type-checked and nothing is silently missing.
  - `createMockAuthContext(overrides?)` — same pattern for `AuthContextType`, covering all 9 fields including the newer password/verification methods.
- No `as any` / double-casts anywhere in either test file.

**2. Added a frontend typecheck step to CI**, mirroring #250's backend fix:
- Added `"typecheck": "tsc --noEmit"` to `package.json`.
- Added a `npm run typecheck` step in `frontend-tests.yml`, running before `npm run test:run`, so a type error in app or test code now fails the PR instead of merging silently.

## Testing

- `npx tsc --noEmit` → clean (was failing with 4 errors before this change)
- `npx vitest run src/__tests__/ForgotPasswordPage.test.tsx src/__tests__/LoginPage.test.tsx` → 6/6 passing
- `npm run test:run` → full suite passing
- Confirmed the pre-existing `eslint` errors on `main` (`react-hooks/set-state-in-effect` in unrelated dashboard/hook files) are untouched by this PR and out of scope

## Summary by CodeRabbit

* **Tests**
  * Improved login and password recovery test coverage with shared, type-safe authentication and routing mocks.
  * Added reusable utilities for configuring test scenarios with custom overrides.
  * Standardized authenticated, unauthenticated, and loading-state test scenarios.

* **Chores**
  * Added an npm type-checking command.
  * Frontend CI now runs type checks before the test suite.
  * Updated workflow permissions for more consistent automated checks.

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -20,4 +20,5 @@ jobs:
           node-version: 22
           cache: npm
       - run: npm ci
+      - run: npm run typecheck
       - run: npm run test:run
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -13,6 +13,8 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "eslint .",
+    "typecheck": "tsc --noEmit",
     "test": "vitest",
     "test:run": "vitest run"
   },
--- a/src/__tests__/ForgotPasswordPage.test.tsx
+++ b/src/__tests__/ForgotPasswordPage.test.tsx
@@ -3,6 +3,7 @@ import { render, screen } from '@testing-library/react';
 import ForgotPasswordPage from '@/app/forgot-password/page';
 import { useAuth } from '@/hooks/useAuth';
 import { useRouter } from 'next/navigation';
+import { createMockRouter } from './test-utils/mockRouter';
 
 vi.mock('next/navigation', () => ({
   useRouter: vi.fn(),
@@ -41,7 +42,7 @@ describe('ForgotPasswordPage', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
-    vi.mocked(useRouter).mockReturnValue({ push: pushMock } as ReturnType<typeof useRouter>);
+    vi.mocked(useRouter).mockReturnValue(createMockRouter({ push: pushMock }));
   });
 
   it('redirects an active logged-in user to /dashboard', () => {
--- a/src/__tests__/LoginPage.test.tsx
+++ b/src/__tests__/LoginPage.test.tsx
@@ -3,13 +3,15 @@ import { render, screen } from '@testing-library/react';
 import LoginPage from '@/app/login/page';
 import { useAuth } from '@/hooks/useAuth';
 import { useRouter } from 'next/navigation';
+import { createMockRouter } from './test-utils/mockRouter';
+import { createMockAuthContext } from './test-utils/mockAuth';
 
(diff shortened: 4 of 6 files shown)


abubakr-itqan (approved):
Reviewed properly - checked out the branch, confirmed the new typecheck script itself passes clean (the four pre-existing errors are actually gone now, not just worked around), ran the full frontend suite (38 files, 220 tests pass), and lint on the touched files is clean. The typed mock factories are a real fix, not a type-check-satisfying shortcut - no `as any`/`as never` anywhere in the final diff, and I like that you caught your own `as never` cast on the second pass and fixed it with a real field instead of hiding it.

Closes the actual gap from #257: frontend tsc now runs in CI before tests, mirroring what #250 did for payload-backend. Also appreciate the unprompted least-privilege permissions fix on the workflow job. Merging.

Issue #228, opened 2026-08-19
[Bug] Comment/report/access-request UI built but not wired into the live app


abubakr-itqan:
## Summary
Comments, resource reporting, access requests, related resources, and a resource preview are fully implemented end-to-end (components + backing use-cases) but never rendered on the live resource detail page. Separately, the dashboard's real approve/deny UI for access requests only appears on mock-data pages wired to `console.log`, while the actually backend-wired requests page renders no action buttons at all. Net effect: none of this works for a real user today, even though almost all of the code exists.

## Screenshot
Live resource detail page (`ratq.itqan.dev/resources/quran-json`) - notice there's no comment section, report button, access-request button, or related resources anywhere on the page:

![Resource detail page missing all interaction UI](https://community.itqan.dev/assets/files/2026-08-19/1787161025-388372-detail-no-interaction-ui.png)

## Background
Confirmed by search: `CommentSection.tsx`, `RelatedResources.tsx`, `AccessRequestButton.tsx`, `ReportButton.tsx`/`ReportModal.tsx`, and `ResourcePreview.tsx` are never imported anywhere under `src/app/**`. `src/app/resources/[slug]/ResourceDetailClient.tsx` (the page that actually renders) only shows `TrustedBySection`, `GithubStatsCard`, and `GithubRepoPreview`.

Separately: `RequestCard.tsx` has working approve/deny buttons with proper `canApproveOrDeny` gating, but it's only used on `src/app/dashboard/requests/page.tsx` and `src/app/dashboard/page.tsx`, both hardcoded with mock arrays - `onApprove`/`onDeny` call `console.log(...)`, not the real API. The page that's actually wired to the backend, `/developer/requests` (using `RequestRow.tsx`), renders status only, with no approve/deny action at all.

## Scope
- [ ] Render `CommentSection`, `ReportButton`/`ReportModal`, `AccessRequestButton`, and `RelatedResources` on `ResourceDetailClient.tsx`
- [ ] Wire real approve/deny actions into the backend-connected `/developer/requests` page (either add buttons to `RequestRow.tsx`, or replace it with the already-built `RequestCard.tsx` and connect its handlers to the real `AccessRequests` API instead of mock data/`console.log`)
- [ ] Guard the comment/report/access-request submission paths so they only activate for `payload`-sourced resources (see the separate resource-id-offset issue - these use `resourceId - 200_000` to recover the Payload doc id, which breaks silently for `cms`-sourced resources)

## Out of Scope
[...]

PR #261, opened 2026-08-25
feat: enhance dashboard and requests pages with dynamic data fetching…

boss-moh:
## Summary by CodeRabbit

* **New Features**
  * Dashboard request lists now display live access requests with loading skeletons and empty states.
  * Developers can approve or deny requests directly, with progress and error feedback.
  * Resource pages now support previews, related resources, comments, access requests, and reporting.
* **Bug Fixes**
  * Access-request updates now connect to the backend and refresh request data.
* **Localization**
  * Added English and Arabic messages for request actions, statuses, loading, and errors.


abubakr-itqan (changes requested):
Hey @boss-moh, reviewed this end to end (checked out the branch, ran tsc-check, ran the full test suite, read through the diff against #228's AC, and traced the actual data flow for the approve/deny path). Good news first: the wiring itself is solid. `ResourceDetailClient` correctly gates the comment/report/access-request/related-resources UI behind `resource.source === 'payload'`, which matches what the issue asked for on the cms-vs-payload split. Both dashboard pages are properly off mock data now. `RequestRow` and its test got cleanly deleted with no dangling references anywhere. tsc-check is clean, and the full suite passes (37 files, 213 tests). I also went and checked the backend access control on the new PATCH specifically, since this is an approve/deny action — `canUpdateAccessRequest` in `AccessRequests.ts` correctly restricts it to the resource's actual publisher or an admin, so that's not just client-side gating, it's enforced server-side too.

Two things I need fixed before this can merge though:

**1. Approve/deny doesn't visibly do anything.** `useUpdateAccessRequest`'s SWR mutation key is `"/api/developer/access-requests"`, but `useDeveloperRequests` actually reads under `['developer', 'requests']` — different key, so the list never revalidates after a decision. The hook also imports `mutate` from `swr` but never calls it. On top of that, `<ToastContainer>` isn't mounted anywhere in the app right now (more on that below, it's not something you introduced), so there's no success/error feedback either. Net effect: the backend write genuinely goes through and persists, but a publisher who clicks Approve sees nothing happen on screen — no toast, no row update. That's the issue's AC #3 ("the decision persists") not being observably true even though it's technically true. Fixing the mutate key (or calling `mutate(['developer','requests'])` on success) should close this.

**2. Needs a test for the new hook.** Per CONTRIBUTING.md: "tests are required, not optional... for a new hook or utility function with any branching logic" — `useUpdateAccessRequest` has onSuccess/onError paths, so it qualifies. Right now this PR actually removes test coverage (`RequestRow.test.tsx`, 72 lines) without adding anything back for the new mutation path.

Two smaller things, not blockers but worth cleaning up while you're in there:
- The new toast strings ("Approving the access request...", etc.) and the `aria-label="Loading access request"` on the skeleton are hardcoded English. CONTRIBUTING.md asks for UI text to go through `messages/{ar,en}.json` instead.
- `dashboard/page.tsx`'s "pending requests" stat is now `requestsResponse.data?.length`, which counts every request regardless of status, not just pending ones. It happened to look right before only because the old mock data was all-pending — with real mixed-status data it'll be wrong.

Now to your three questions:

**Toast messages not appearing** — tracked it down, and it's not something wrong with your code specifically. `<ToastContainer>` isn't rendered anywhere in the app at all right now, in any layout. `useToast()` as it's written creates its own local, per-component-instance toast queue rather than a shared context (you can see the same pattern already in `ReportModal.tsx`), so even where it's called, nothing is actually listening for it to display. This predates your PR — worth its own issue if you want to file one, since it affects every `toast()` call in the codebase, not just this page.

**Client-side rendering vs SSR** — fair question, but these pages read the authenticated user from a client-side token (localStorage), so moving them to the server would mean building out real server-side auth first, that's a bigger separate change than this issue's scope.

**RequestCard managing its own request/status internally** — I get the appeal, especially for the loading-state ergonomics. The reason it's structured with the parent owning the fetch is that both `/dashboard` and `/developer/requests` need to render the same list with consistent counts and loading state, so a single shared fetch avoids two independent components double-fetching and possibly showing different data for a second. If this needs to scale to more places doing their own thing, that's a reasonable time to revisit it, so it's a good instinct.

One process note: this branch is named `bug/comment`, and this repo's Branch Naming Check only allows `feature/`, `feat/`, `hotfix/`, `community/`, `chore/`, `release/`, `dependabot/` — so once the above is sorted, merging will need an admin override on that check. Not something you need to redo the branch for, just flagging why you'll see it red.

abubakr-itqan:
@boss-moh just checking in on this one - no rush at all, just making sure the review didn't slip past you. Let me know if you're still working through the two blocking items or if you'd like a hand with anything.

boss-moh:
> @boss-moh just checking in on this one - no rush at all, just making sure the review didn't slip past you. Let me know if you're still working through the two blocking items or if you'd like a hand with anything.

sorry for the late response 
there was problem in the internet service providers  therefore I was not able to send or reply 

I have already finish the work

boss-moh:
> Hey @boss-moh, reviewed this end to end (checked out the branch, ran tsc-check, ran the full test suite, read through the diff against #228's AC, and traced the actual data flow for the approve/deny path). Good news first: the wiring itself is solid. `ResourceDetailClient` correctly gates the comment/report/access-request/related-resources UI behind `resource.source === 'payload'`, which matches what the issue asked for on the cms-vs-payload split. Both dashboard pages are properly off mock data now. `RequestRow` and its test got cleanly deleted with no dangling references anywhere. tsc-check is clean, and the full suite passes (37 files, 213 tests). I also went and checked the backend access control on the new PATCH specifically, since this is an approve/deny action — `canUpdateAccessRequest` in `AccessRequests.ts` correctly restricts it to the resource's actual publisher or an admin, so that's not just client-side gating, it's enforced server-side too.
> 
> Two things I need fixed before this can merge though:
> 
> **1. Approve/deny doesn't visibly do anything.** `useUpdateAccessRequest`'s SWR mutation key is `"/api/developer/access-requests"`, but `useDeveloperRequests` actually reads under `['developer', 'requests']` — different key, so the list never revalidates after a decision. The hook also imports `mutate` from `swr` but never calls it. On top of that, `<ToastContainer>` isn't mounted anywhere in the app right now (more on that below, it's not something you introduced), so there's no success/error feedback either. Net effect: the backend write genuinely goes through and persists, but a publisher who clicks Approve sees nothing happen on screen — no toast, no row update. That's the issue's AC #3 ("the decision persists") not being observably true even though it's technically true. Fixing the mutate key (or calling `mutate(['developer','requests'])` on success) should close this.
> 
> **2. Needs a test for the new hook.** Per CONTRIBUTING.md: "tests are required, not optional... for a new hook or utility function with any branching logic" — `useUpdateAccessRequest` has onSuccess/onError paths, so it qualifies. Right now this PR actually removes test coverage (`RequestRow.test.tsx`, 72 lines) without adding anything back for the new mutation path.
> 
> Two smaller things, not blockers but worth cleaning up while you're in there:
> 
> * The new toast strings ("Approving the access request...", etc.) and the `aria-label="Loading access request"` on the skeleton are hardcoded English. CONTRIBUTING.md asks for UI text to go through `messages/{ar,en}.json` instead.
> * `dashboard/page.tsx`'s "pending requests" stat is now `requestsResponse.data?.length`, which counts every request regardless of status, not just pending ones. It happened to look right before only because the old mock data was all-pending — with real mixed-status data it'll be wrong.
> 
> Now to your three questions:
> 
> **Toast messages not appearing** — tracked it down, and it's not something wrong with your code specifically. `<ToastContainer>` isn't rendered anywhere in the app at all right now, in any layout. `useToast()` as it's written creates its own local, per-component-instance toast queue rather than a shared context (you can see the same pattern already in `ReportModal.tsx`), so even where it's called, nothing is actually listening for it to display. This predates your PR — worth its own issue if you want to file one, since it affects every `toast()` call in the codebase, not just this page.
> 
> **Client-side rendering vs SSR** — fair question, but these pages read the authenticated user from a client-side token (localStorage), so moving them to the server would mean building out real server-side auth first, that's a bigger separate change than this issue's scope.
> 
> **RequestCard managing its own request/status internally** — I get the appeal, especially for the loading-state ergonomics. The reason it's structured with the parent owning the fetch is that both `/dashboard` and `/developer/requests` need to render the same list with consistent counts and loading state, so a single shared fetch avoids two independent components double-fetching and possibly showing different data for a second. If this needs to scale to more places doing their own thing, that's a reasonable time to revisit it, so it's a good instinct.
> 
> One process note: this branch is named `bug/comment`, and this repo's Branch Naming Check only allows `feature/`, `feat/`, `hotfix/`, `community/`, `chore/`, `release/`, `dependabot/` — so once the above is sorted, merging will need an admin override on that check. Not something you need to redo the branch for, just flagging why you'll see it red.

At client-side rendering vs SSR 
- there are some better solutions  for auth (some libs supporting both side and client rendering for auth ) 
- Also it is not safe  to use local storge

abubakr-itqan (approved):
Re-reviewed after the update - checked out the new commits, ran tsc-check, the full test suite (all green), and read through the diff again.

Both blocking items are properly closed:
1. Approve/deny now visibly works - RequestCard owns its own mutation via useUpdateAccessRequest, which shares the exact same SWR key (['developer', 'requests']) as useDeveloperRequests, so triggering the mutation revalidates the list automatically. No manual mutate() call needed, this is the cleaner fix. It also shows a loading state on the button and an inline error message on failure, which is a real improvement over what I originally suggested.
2. useUpdateAccessRequest.test.tsx is a solid addition - covers the approve path, deny path, and the error path with a real assertion on the returned data.

Nice touch keeping the shared fetch in the parent while moving the mutation logic into RequestCard itself - that's a good middle ground between your original question and the constraint I mentioned.

On the SSR/localStorage point: fair concern in general, but that's a bigger auth-architecture change than this PR's scope, worth its own discussion if you want to raise it separately.

This is good to merge. Branch naming check will still show red (bug/comment isn't on the allow-list), that's expected and I'll override it on merge.

PR #264, opened 2026-08-26
feat(oauth): exchange one-time code for session token

### LINKED ISSUE #229: [Security] OAuth callback passes session token in URL query string
## Summary
The GitHub OAuth callback redirects the browser to the frontend with the live session token in the URL's query string. URL-borne tokens land in browser history, server access logs, and any `Referer` header sent by whatever the callback page navigates to next - a standard token-leak vector.

## Background
`payload-backend/src/app/oauth/github/callback/route.ts` redirects with `NextResponse.redirect(\`${frontendUrl}/oauth/callback?token=${token}\`)`. The token is a live, usable session credential at the moment it's placed in the URL.

## Scope
- [ ] Stop passing the session token via URL query string in the OAuth callback redirect
- [ ] Replace with an approach that doesn't expose the token in URLs/logs - e.g. a short-lived one-time exchange code in the URL that the frontend immediately swaps for the real token via a POST request, or setting the token via a response header/cookie on the redirect itself
- [ ] Confirm the frontend's `/oauth/callback` page is updated to match whatever mechanism replaces the query param

## Out of Scope
- Broader auth/token-storage hardening (localStorage token storage is a separate, already-known tradeoff, not this issue's scope)

## Technical Notes
- `payload-backend/src/app/oauth/github/callback/route.ts` - where the redirect happens
- Frontend `/oauth/callback` page - consumes the token today, needs to match the new mechanism

## Acceptance Criteria
- [ ] The session token no longer appears anywhere in a URL during the OAuth login flow
- [ ] GitHub OAuth login still works end-to-end after the change

## Difficulty
Small-Medium - contained to the OAuth callback route and its matching frontend page.

husamemadH:
Closes #229

## The problem

The GitHub OAuth callback redirected to the frontend with the session token in the query string:

```ts
NextResponse.redirect(`${frontendUrl}/oauth/callback?token=${token}`)
```

That token is a live session credential with a multi-day lifetime, so putting it in a URL writes it to
browser history, the reverse proxy's access log, and any `Referer` header the callback page sends.
Anyone who can read those has a working session for that user, with no exploit involved.

## The approach

Keep the session token out of the URL entirely by splitting the handoff into two steps, the same
pattern OAuth itself uses one hop earlier:

1. The callback signs a **60-second, single-use code** and puts it in the redirect **fragment**.
   Fragments are never sent to a server and never appear in a `Referer`, so even this short-lived
   value stays out of access logs.
2. The frontend POSTs that code to a new `/oauth/github/exchange`, which returns the session token in
   the **response body** - a channel nothing logs, caches or forwards.

### Flow

Before:

```
GET  /oauth/github/callback   307 -> /oauth/callback?token=eyJ...   <- credential in the URL
GET  /oauth/callback?token=   200
```

After:

```
GET  /oauth/github/callback   307 -> /oauth/callback#code=eyJ...    <- 60s single-use ticket
GET  /oauth/callback          200                                   <- no credential in the URL
POST /oauth/github/exchange   200 -> { "token": "eyJ..." }          <- credential in the body
```

### How single-use works without a new table

The obvious design is a table of spent codes. That isn't needed here, because a row already gets
created at exactly the right moment.

Payload rejects any session token whose `sid` has no matching row in the user's `sessions` array, so
today's callback creates that row and signs the token together. This PR splits those two moments: the
callback picks a `sid` but writes nothing, and the exchange endpoint creates the row. So "has this code
been redeemed?" becomes "does a session with this sid already exist?" - and creating the session *is*
spending the code.

A side effect worth having: an abandoned login now leaves nothing behind, where previously it left an
orphan session row until it expired.

### Why exchange codes use a separate signing key

They're signed with `sha256(payload.secret + ':oauth-exchange')` rather than `payload.secret`. If both
used the same key, an exchange code would be a structurally valid session token in Payload's eyes, and
only a claim check would stand between the two. Different keys make the confusion impossible, and two
tests pin that in both directions.

## Changes

Backend:
- `lib/oauth/github.ts` - `signExchangeToken` / `verifyExchangeToken`
- `oauth/github/callback/route.ts` - signs a code instead of a session token, redirects with `#code=`,
  no longer creates the session
- `oauth/github/exchange/route.ts` (new) - verifies the code, rejects replays, creates the session,
  returns the token. Includes its own CORS/`OPTIONS` handling, since `/oauth/*` sits outside Payload's
  `/api` catch-all and so gets none of `config.cors` applied for it. The allow-list is read off
  `payload.config.cors` to keep one source of truth.

Frontend:
- `oauth/callback/page.tsx` - reads the code from `location.hash` (fragments are invisible to
  `useSearchParams`), strips it from the address bar immediately, then redeems it
- `payload-auth-repository.ts` - `exchangeOAuthCode()`
- `complete-oauth.ts` - exchange, then the unchanged `loginWithToken`
- `useAuth.tsx` - `loginWithToken` -> `loginWithCode`, since it takes a code now. This is what touches
  the 8 test files in the diff (a renamed mock property). Happy to revert the rename if you'd rather
  keep the diff smaller.

Everything downstream is untouched: the token is still validated via `/api/users/me`, still stored by
`setAuthTokens`, still sent as `Authorization: JWT`.

## Testing

`payload-backend` - 9 new tests in `lib/oauth/github.test.ts`: round-trip, 60s expiry, wrong secret,
tampered token, wrong purpose, expired, missing claims, and both directions of the key separation.

Frontend - 5 new tests in `OAuthCallbackPage.test.tsx`: reads the code from the fragment, strips it
from the address bar, redeems only once on a double render, and both failure redirects.

Before - the session token sits in the callback URL:

After - the same row, with no credential in it:

Verified by hand against a local stack with a real GitHub OAuth app:
- login works end to end, no `token=` anywhere in the flow, and browser history holds only a bare
  `/oauth/callback`
- POSTing the same code twice returns `200` then `401 code_already_used`
- a session token POSTed as a code is rejected
- tampered, malformed and expired codes are rejected
- a disallowed `Origin` gets no `Access-Control-Allow-Origin` header

Note: `ResourceDetailPage.test.tsx` has 4 pre-existing failures on `main` (an `AbortSignal` mock issue),
unrelated to this change.

## Concurrent redemption

The first version of this PR checked for an existing session and then wrote one, which are two separate
operations - two requests carrying the same code could both pass the check before either wrote. That is
now a single conditional INSERT: `users_sessions.id` is the primary key, so `ON CONFLICT (id) DO NOTHING`
makes Postgres the arbiter, and an empty `RETURNING` is how the losing request finds out.

Verified with 12 simultaneous redemptions of one code: one `200`, eleven `401 code_already_used`.

This is asserted by a schema constraint rather than by a unit test - `payload-backend`'s suite is pure
unit tests with no database, so a concurrency test would need a Postgres-backed integration setup that
doesn't exist in this repo yet. Happy to add one if you'd like that infrastructure.

## Summary by CodeRabbit

* **New Features**
  * Improved GitHub OAuth security with short-lived, single-use authorization codes.
  * OAuth callbacks now use URL fragments and securely exchange codes for sessions.
  * Added validation for invalid, expired, reused, or tampered authorization codes.
  * Added origin validation, CORS handling, and no-cache responses for exchanges.

* **Bug Fixes**
  * Prevented OAuth credentials from appearing in query-string URLs or browser history.

* **Tests**
  * Added coverage for code validation, redemption, expiration, redirects, and callback behavior.


abubakr-itqan (changes requested):
This is genuinely excellent work, shukran for the thorough writeup and the honesty about the known limitation. The fragment+exchange design is exactly right for #229 - checked out the branch, tsc-check and the full suite (frontend + payload-backend) both pass clean, and the 9 new github.test.ts cases actually pin the security-relevant behavior (key separation both directions, expiry, tampering, wrong purpose). Read the diff end to end and it's clean - no safety concerns, no scope creep.

One thing I'd like closed before this merges: the race you flagged yourself. `sessions.some((session) => session.id === claims.sid)` followed by `payload.update(...)` is a read-then-write, so two concurrent POSTs with the same code can both pass the check and both create a session - which means two live sessions from one exchange code instead of the single-use guarantee the whole design is built around. I know the practical window is narrow (60s, fragment-only, single recipient), but since #229 is specifically about tightening this auth surface, I'd rather close it properly than accept a residual gap on the same flow we just hardened.

You already named the fix - a codes/tickets collection with a unique constraint on `sid` (or a Payload  guard on the update, if Postgres can enforce uniqueness without a full new collection) would make redemption atomic instead of check-then-act. Happy to look at whichever shape you land on. Everything else here is ready to go the moment that's in.

abubakr-itqan:
@husamemadH just checking in on this one - no rush at all, just making sure the review didn't slip past you. Let me know if you're still working through the atomic-fix piece or if you'd like a hand with anything.

husamemadH:
yeah i have this issue queued for my tasks today inshallah today i will work on the follow up

abubakr-itqan (approved):
Re-reviewed after the update - checked out the new commit, ran tsc-check and both test suites (frontend + payload-backend), all clean.

This closes the race properly. Rather than a read-then-write check, the exchange route now does a single conditional INSERT against users_sessions with ON CONFLICT (id) DO NOTHING, letting Postgres's own primary-key constraint be the arbiter - two concurrent redemptions of the same code can only ever produce one successful insert, and the loser gets an empty RETURNING to know it lost. That's exactly the atomic guarantee I was asking for, and it didn't need a new migration or collection to get there.

Good catch also pruning expired sessions separately now, since the targeted insert doesn't do that as a side effect the way the old full-array payload.update did.

This is good to merge.

PR #265, opened 2026-08-26
Fix auth error cleanup across authentication routes

### LINKED ISSUE #246: [Bug] Authentication errors persist when navigating between login and registration
## Summary

Authentication error messages persist when navigating between the login and registration pages and are only cleared after a hard refresh.

For example, after submitting invalid login credentials, the message:

`The email or password provided is incorrect.`

continues to appear when navigating from the login page to the registration page.

The stale error can also remain when navigating back to login.

Authentication errors should be scoped to the relevant authentication flow and cleared when the user intentionally navigates between login and registration.

## Background

This was initially noticed while implementing session-expiry handling for #165.

A session-expired message displayed on `/login` was also appearing after navigating to the registration page. At first, this appeared to be specific to the new session-expiry behavior.

Further testing showed that the same behavior already occurs with regular login errors.

For example:

1. Navigate to `/login`.
2. Enter invalid credentials.
3. Submit the login form.
4. Confirm that the following error appears:
   `The email or password provided is incorrect.`
5. Navigate to the registration page.
6. The previous login error remains visible.
7. Navigate back to login.
8. The stale error may remain until the page is hard-refreshed.

This confirms that the problem is not specific to session expiry. It is a broader authentication error-state lifecycle issue.

## Current Behavior

Authentication errors can survive navigation between the login and registration flows.

For example:

```text
Invalid login attempt
        ↓
"The email or password provided is incorrect."
        ↓
Navigate to Register
        ↓
Stale login error is still displayed
        ↓
Navigate back to Login
        ↓
Error remains
        ↓
Hard refresh
        ↓
Error is cleared
```

Session-expiry messages may exhibit the same behavior because they use the same authentication UI/error flow.

## Desired Behavior

Authentication errors should only remain visible while they are relevant to the current authentication flow.

Navigating intentionally between login and registration should clear stale authentication errors.

For example:

```text
Invalid login attempt
        ↓
Login error displayed
        ↓
Navigate to Register
        ↓
Error cleared
        ↓
Registration form starts with clean state
```

Likewise, a session-expired message shown after redirecting a user to `/login` should explain why they were redirected, but should not continue appearing after the user leaves the login flow.

## Why This Is Separate From #165

#165 specifically addresses expired or invalid authenticated sessions during API requests:

* detecting when an existing session is no longer valid
* distinguishing session expiry from genuine permission errors
* clearing stale authentication state
* redirecting the affected user to `/login`
* displaying an appropriate session-expired message

While implementing that behavior, the persistent-error problem became more visible because the new `session_expired` message could also survive navigation to registration.

However, testing with a normal invalid-login error reproduced the same issue:

`The email or password provided is incorrect.`

This demonstrates that stale authentication errors are not caused by the session-expiry implementation and are not limited to expired sessions.

Fixing the general authentication error lifecycle inside #165 would therefore expand that issue beyond its intended session-expiry scope.

This issue tracks the broader login/register error-state behavior separately so that #165 can remain focused on session-expiry detection and handling.

## Scope

* [ ] Clear stale login errors when navigating from login to registration
* [ ] Clear stale registration errors when navigating from registration to login
* [ ] Ensure session-expired messages do not persist onto the registration flow
* [ ] Ensure a newly opened authentication form starts without errors belonging to the previous form
* [ ] Preserve errors while the user remains on the form where the error occurred
* [ ] Add or update tests covering authentication error cleanup during navigation

## Out of Scope

* Session-expiry detection from API responses — tracked in #165
* Token expiry or token lifetime changes
* Refresh-token/session-renewal mechanisms
* Changes to authentication validation rules
* Changes to the wording of existing authentication errors
* Redesigning the login or registration forms

## Technical Notes

The underlying issue appears to be related to authentication error state surviving client-side navigation rather than being scoped or reset when switching authentication flows.

The exact implementation should be investigated rather than assuming where the reset belongs.

Potential areas to inspect include:

* shared authentication state used by login and registration
* login and registration form lifecycle
* error state cleanup when an authentication component unmounts
* navigation between `/login` and `/register`
* handling of session-expiry messages introduced as part of #165

The solution should avoid clearing an error prematurely while the user remains on the same form. The goal is specifically to prevent an error from one authentication context leaking into another.

## Acceptance Criteria

* [ ] Invalid login credentials display the appropriate login error
* [ ] Navigating from login to registration clears the previous login error
* [ ] Registration opens without displaying an error originating from the login form
* [ ] Navigating from registration back to login does not restore the stale previous error
* [ ] A session-expired message shown on login does not appear on registration after navigation
* [ ] Authentication errors continue to display normally while the user remains on the form that produced them
* [ ] A hard refresh is no longer required to clear stale authentication errors
* [ ] Existing login and registration behavior continues to work

## References

Related: #165 — expired-session detection and re-login handling.

This issue was discovered while working on #165 but is intentionally tracked separately because it can be reproduced independently using normal invalid login credentials.

## Difficulty

Beginner / Easy

## Good First Issue?

Yes — the behavior is reproducible through normal authentication navigation and appears limited to UI/error-state lifecycle management rather than authentication protocol or backend changes.

## Contributor Checklist

* [ ] Read `CONTRIBUTING.md`
* [ ] Keep changes focused on authentication error-state cleanup
* [ ] Add or update tests where practical
* [ ] Verify both login → register and register → login navigation
* [ ] Ensure `tsc-check`/lint/tests pass before opening a PR

Ramahadam:
## Summary

Fixes authentication errors persisting across client-side navigation between authentication flows.

The shared `AuthProvider.error` state previously survived route changes, which allowed errors from one flow to appear in another. For example, a failed login error could appear on Forgot Password, and registration errors could later appear on Login.

This change adds explicit auth-error cleanup when entering a new auth route.

Login preserves `SESSION_EXPIRED_REASON` so users redirected after session expiry still see the reason for the redirect.

Fixes #246

## Changes

* add a stable `clearError` API to `AuthProvider`
* clear stale auth errors when entering Login, Register, and Forgot Password
* preserve the session-expired reason when entering Login
* perform cleanup at the page boundary so LoginForm loading/remount cycles do not clear newly produced login errors
* update existing auth mocks for the new context API
* add regression coverage for auth error lifecycle across navigation

## Testing

Added `AuthErrorNavigation.test.tsx` covering:

* registration error does not leak into Login
* login error does not leak into Forgot Password
* failed login error survives the LoginForm loading/remount cycle
* old login errors do not reappear after leaving Login and returning
* session-expired messages remain visible on Login but do not leak into another auth flow

Relevant test results:

* `AuthErrorNavigation.test.tsx`: 5 passed
* `useAuth.test.tsx`: 11 passed
* Login/Register/Forgot Password/Reset Password form tests: 11 passed
* `npx tsc --noEmit`: passed

`npm run lint` still reports pre-existing errors in unrelated dashboard/resource files; no lint errors from this change were identified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale authentication errors from persisting when moving between login, registration, and password-reset pages.
  * Preserved session-expiration messages while clearing unrelated authentication errors.
  * Improved error-state handling when forms are remounted or revisited.

* **Tests**
  * Added comprehensive coverage for authentication navigation, error clearing, and session-expiration messaging.
  * Updated authentication form and page tests to reflect improved error handling.


abubakr-itqan (approved):
Reviewed this end to end - checked out the branch, ran tsc-check and the full test suite (both clean), and read through the diff. The fix is exactly right: `clearError` on the auth context with a `preserveSessionExpiry` option, cleared at the page boundary (login/register/forgot-password) rather than inside the form components, which is what avoids the LoginForm remount/loading cycle clobbering a fresh error. The new `AuthErrorNavigation.test.tsx` covers the actual scenarios that mattered here (error leaking across routes, session-expired message surviving, old errors not reappearing). No lint issues in the touched files. This closes out #246 properly - nice work getting through it.

Issue #262, opened 2026-08-26
[Bug] Toast notifications never display - ToastContainer is never mounted


abubakr-itqan:
## Summary
`toast()` calls across the app are silent no-ops. `useToast()` returns local, per-component-instance state (`toasts`, `removeToast`), and the `ToastContainer` component that actually renders toasts on screen is never mounted anywhere in the app - not in any layout, not even by the components that call `toast()`.

## Background
Confirmed in `src/shared/ui/Toast.tsx`: `useToast()` is a hook that owns its own `toasts` state and returns `{ toast, toasts, removeToast }` - it's not a shared context, so every component that calls it gets an independent, isolated toast queue. `ToastContainer` is a separate component that takes `toasts`/`removeToast` as props and renders the actual UI.

The only current caller, `src/modules/resources/components/ReportModal.tsx`, destructures just `const { toast } = useToast();` - it calls `toast()` on submit but never renders `<ToastContainer toasts={toasts} removeToast={removeToast} />`, so the call updates state nobody reads. No `<ToastContainer>` is rendered in any root layout either, so there's no app-wide fallback catching this.

Net effect: any feature that calls `toast()` for user feedback (success/error messages) - existing or new - shows nothing on screen. This surfaced concretely in PR #261 (wiring up publisher approve/deny), where the approve/deny action persists correctly server-side but gives the user zero visual confirmation.

## Scope
- [ ] Turn toast state into a single shared source (e.g. React context provider) instead of a hook-per-caller
- [ ] Mount one `ToastContainer` at the app root (root layout) so it's available everywhere without every caller needing to render it itself
- [ ] Update `ReportModal.tsx` (and any other caller) to the new pattern

## Out of Scope
- New toast types/styles - the visual design already exists and works, this is purely about making it reachable
- Fixing other callers that assume `toast()` works (those get fixed for free once this lands, no separate follow-up needed)

## Technical Notes
- `src/shared/ui/Toast.tsx` - `useToast()` hook and `ToastContainer` component
- `src/modules/resources/components/ReportModal.tsx` - current (broken) usage pattern
- Root layout(s) under `src/app/` - where a shared `ToastContainer` needs to be mounted once

## Acceptance Criteria
- [ ] Calling `toast()` from any component shows a real toast on screen
- [ ] Two different components calling `toast()` around the same time both show up correctly, not just the last one
- [ ] `ReportModal.tsx`'s existing toast call now actually displays

## Difficulty
[...]

PR #267, opened 2026-08-28
Feat/toast bug

malakmuayad11:
## Summary
Solved the issue of toast messages not being displayed on the screen by using a shared context: `ToastContext`. The `ToastProvider` is used in the root layout, so all toast messages are shared between the app's components. This closes issue #262.

## Changes

- Created a `ToastContext` and `ToastProvider` in `src/shared/ui/Toast.tsx`. I tried to match the existing architecture used in the project, like `LanguageContext` and `AuthContext`.
- `ToastProvider` is mounted once in the root layout, which also ensures that `ToastContainer` is rendered only once, as required.
- Updated `useToast` to consume `ToastContext` and handle the null-context case. Existing usage in `ReportModal` did not need to change.
- Updated `ReportModal.test.tsx` to wrap the component with `ToastProvider`, preventing null-context errors during tests.

## Summary by CodeRabbit

* **Improvements**
  * Updated toast notifications to use a shared application-wide provider.
  * Toast messages are now rendered consistently across the header, main content, and footer.

* **Tests**
  * Updated report modal tests to support the shared toast notification system.
  * Preserved English and Arabic placeholder coverage.

### DIFF
--- a/src/__tests__/ReportModal.test.tsx
+++ b/src/__tests__/ReportModal.test.tsx
@@ -1,38 +1,49 @@
-import { act } from 'react';
-import { render, screen } from '@testing-library/react';
-import { describe, expect, it } from 'vitest';
-import { ReportModal } from '@/modules/resources/components/ReportModal';
-import { LanguageProvider } from '@/shared/ui/i18n/LanguageContext';
+import { act } from "react";
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { ReportModal } from "@/modules/resources/components/ReportModal";
+import { LanguageProvider } from "@/shared/ui/i18n/LanguageContext";
+import { ToastProvider } from "@/shared/ui/Toast";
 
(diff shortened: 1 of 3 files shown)


abubakr-itqan (approved):
Reviewed end to end - checked out the branch, ran tsc-check and the full test suite (both clean), read through the diff, and specifically scanned the whole codebase for every caller of useToast to make sure the context-shape change doesn't break anything else. ReportModal is genuinely the only consumer, and it already only destructured { toast }, so this lands with zero call-site changes needed there.

The fix itself is the right one: a shared ToastContext/ToastProvider mounted once at the root layout, matching the existing LanguageContext/AuthContext pattern in this codebase - exactly what the old per-component-instance useToast was missing. ToastContainer now genuinely has something listening to it. Good catch wrapping the ReportModal test in ToastProvider too, since useToast now throws outside a provider instead of silently doing nothing.

This closes out #262 for real. Merging.

PR #268, opened 2026-08-28
fix: generate OAuth throwaway password within validator limits (#263)

### LINKED ISSUE #263: [Bug] GitHub OAuth signup 500s for any new user - generated password exceeds the 64-char limit
Alsalam alykom,

While working on #229 I hit a bug in the GitHub OAuth flow that looks like it was introduced by #259.
It only affects users signing in with GitHub for the **first time** - anyone who already has an account
is unaffected, which is probably why it hasn't been noticed.

## What's wrong

Signing in with GitHub using an email that has no RATQ account yet fails with a blank
`HTTP ERROR 500` on `api.ratq.itqan.dev`. The user is never created and never logged in.

Expected: a new account is created and the user lands on the dashboard.

Reproduced on `beta.ratq.itqan.dev` and locally on `main` (c47a427).

## Where

`payload-backend/src/app/oauth/github/callback/route.ts:93`

The callback creates the new user with a throwaway password:

```ts
password: crypto.randomUUID() + crypto.randomUUID(),
```

A UUID is 36 characters, so this is **72**. #259 added `passwordValidation` as a `beforeValidate`
hook on the Users collection (`payload-backend/src/collections/Users.ts:17`), which rejects anything
over 64 characters:

```
⨯ Error [APIError]: Password must be between 8 and 64 characters long
    at passwordValidation (src/collections/hooks/validate-password.ts:12:15)
    at async GET (src/app/oauth/github/callback/route.ts:93:7)
```

The route has no `try`/`catch`, so the 422 escapes as a bare 500 instead of the usual
`/login?error=...` redirect the route uses for its other failure paths.

Why existing users are unaffected: that branch only runs for accounts that don't exist yet. Returning
users take the `payload.update` path at `route.ts:120`, which only writes `sessions` and sends no
password, and the hook returns early when `password` is undefined (`validate-password.ts:8`).

## Steps to reproduce

1. Make sure the GitHub account's verified email has no existing RATQ account
2. Go to https://beta.ratq.itqan.dev/login
3. Click "Sign in with GitHub" and approve
4. GitHub redirects back to `api.ratq.itqan.dev/oauth/github/callback?code=...` -> `HTTP ERROR 500`

## Suggested fix

Generate a password that fits the limit, e.g.:

```ts
password: randomBytes(32).toString('hex'),   // exactly 64 chars
```

This is the same generator already used for API keys in
`payload-backend/src/collections/APIKeys.ts:15`, and it's stronger than what it replaces
(256 bits vs 122 for a UUIDv4). The value is never usable for login - these accounts only ever
authenticate through GitHub - it only exists because Payload requires a password on `auth: true`
collections.

I've verified this locally: with the change, first-time GitHub signup creates the user and logs in
normally.

## Acceptance criteria

- [ ] First-time GitHub sign-in creates the account and logs the user in
- [ ] Existing users can still sign in with GitHub
- [ ] A regression test covers the generated password being within the validator's limits

Happy to open a PR for this if you'd like - just let me know before I start, since it touches the same
file as #229 and I'd keep it as a separate PR.

Walid-Khalfa:
**Fixes #263**

First-time GitHub OAuth signups fail with a bare HTTP 500 because the callback route creates new users with crypto.randomUUID() + crypto.randomUUID() (72 chars), which exceeds the 64-char maximum enforced by the passwordValidation hook (added in #259, Users.ts:17). Since the route has no try/catch, the validation error bubbles up as an unhandled 500 instead of a redirect.

**Changes**
Replace crypto.randomUUID() + crypto.randomUUID() with crypto.randomBytes(32).toString('hex') (exactly 64 hex chars) in payload-backend/src/app/oauth/github/callback/route.ts
Matches the existing pattern used in APIKeys.ts:15 (as suggested in the issue)
Add regression test route.test.ts asserting the generated password respects the 8-64 char constraint enforced by passwordValidation

**Acceptance criteria (from #263)**
 First-time GitHub sign-in creates the account and logs the user in (verified via regression test mocking the full OAuth flow)
 Existing users can still sign in with GitHub (test covers the payload.update path — no create call, no password sent)
 Regression test covers the generated password being within the validator's limits

**Testing**
 npm test passes (96 tests, including new regression test)
 npm run lint — no new errors on changed files
 Regression test verified: fails on old code (72 > 64), passes on new code (64 chars)

**Notes**
Google OAuth: verified there is no Google OAuth callback route in the codebase — only GitHub is affected (cc @ekafe re: your comment on the issue)
The route still lacks a try/catch around payload.create() (mentioned in the issue as a secondary concern) — out of scope for this fix, but happy to open a follow-up issue if useful

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GitHub sign-ups failing when automatically generated passwords exceeded the maximum allowed length.
  * New accounts now receive valid passwords that comply with password requirements.
  * Existing GitHub users continue to sign in without creating duplicate accounts.

* **Tests**
  * Added coverage for first-time GitHub sign-ups and returning-user logins.


Walid-Khalfa:
Salam @abubakr-itqan 

Ready for review.

Fix: throwaway password now 64 hex chars (crypto.randomBytes(32)), within the passwordValidation limit (#259). Fixes #263.
Tests: regression test added , full suite passes (96 tests), lint clean.
CodeRabbit: latest review clean. The older comment is stale (previous commit, removed via force-push).
Out of scope (noted in #263): no try/catch around payload.create(), happy to open a follow-up issue if wanted.

Thanks! 🙏

abubakr-itqan:
Reviewed this - the fix itself is solid (checked out the branch, ran tsc-check and the full backend test suite, both clean; the regression test genuinely fails on the old code and passes on the new one). No safety concerns.

One thing before this can merge: #264 just landed and it touches this exact same file (the OAuth callback route), right next to where your password fix sits - it removes the session-creation block that used to sit just after the password line. That's created a real conflict, not a false one.

Could you rebase onto the latest main and push? Your change itself is tiny (just the crypto.randomBytes line and the new import), so it should resolve cleanly - you're just keeping your fix in the file's new shape. Happy to take another look the moment it's through.

abubakr-itqan:
Correction on my last comment - I misread the merge-tree output. GitHub's own mergeability check (and a local test merge I just ran) both confirm this actually merges cleanly against main as-is, no rebase needed. Sorry for the noise - proceeding to merge this now.

abubakr-itqan:
Update: my correction above was half right - the text does merge cleanly (no rebase needed), but the *combined* result actually breaks 2 tests in your own route.test.ts, since #229's fix (merged as #264 right after your PR opened) changed the callback route to call `signExchangeToken` instead of `signSessionJWT` directly.

Your test's mock only stubs `pickVerifiedEmail` and `signSessionJWT`:
```ts
vi.mock('@/lib/oauth/github', () => ({
  pickVerifiedEmail: vi.fn(),
  signSessionJWT: vi.fn(),
}))
```
Once merged with current main, the route calls the now-unmocked `signExchangeToken`, so both tests fail with "No signExchangeToken export is defined on the mock."

Fix is small - add it to the mock, e.g.:
```ts
vi.mock('@/lib/oauth/github', () => ({
  pickVerifiedEmail: vi.fn(),
  signSessionJWT: vi.fn(),
  signExchangeToken: vi.fn().mockResolvedValue('fake-exchange-token'),
}))
```
Merge main into your branch first so you're testing against the same code that'll actually run, then adjust the mock. Sorry for the back-and-forth on this one.

Walid-Khalfa:
Salam @abubakr-itqan , I Updated the branch against current upstream and aligned the regression test with the new signExchangeToken flow. CI is green now, with the full backend suite passing 171/171 tests and typecheck clean. Thanks for the clarification.

abubakr-itqan (approved):
Reviewed the fix - matches #263's own suggested approach exactly (crypto.randomBytes(32).toString('hex')), same pattern already used in APIKeys.ts. Ran the full backend suite (171/171 passing) and independently swapped in the old route.ts to confirm the new regression test actually fails on the pre-fix code and passes on the new one. tsc and lint clean on the touched files. This is a strict entropy improvement too (256-bit vs the old 122-bit from two concatenated UUIDs), still GitHub-only auth so no live risk either way. Merging.

Issue #158, opened 2026-08-02
[Bug] Invite-by-email form and its neighboring dead access-requests stub


abubakr-itqan:
> **Note (2026-08-03):** file paths below were updated to match the `src/` DDD restructure that landed on `beta` - the described issue itself is unaffected, only locations changed.

# Decide the fate of the disabled "invite developer by email" form, and wire up its neighboring dead stub

## Summary
A publisher managing an API-type resource sees two things that don't work: a fully disabled "invite by email" form, and a hardcoded "no pending access requests" message that never reflects reality - both in the same tab, both explicitly marked as placeholders in the code. The second one is the easier fix, since the real `AccessRequests` collection already exists and could back it directly.

## Background
`src/app/developer/resources/[slug]/page.tsx`'s "Access" tab (shown only when `resource.type === 'api'`) contains two placeholders:
1. An invite-by-email form (email input, scope select, "دعوة"/Invite button) - all three controls `disabled`, no handlers, with a `TODO: connect to backend API` comment.
2. Immediately below it, a hardcoded "لا توجد طلبات وصول معلقة" (no pending access requests) message with an explicit `Access requests list placeholder` comment - this always shows the same text regardless of whether the resource actually has pending requests.

There's a corresponding mock function, `api.developer.access.inviteByEmail`, but the form doesn't even call it. The access-requests-list placeholder, unlike the invite form, has no decision to make - `AccessRequests` is a real, working collection; this is just an unwired list.

**Current behavior:** Both pieces are non-functional placeholders regardless of actual account state.
**Desired behavior:** The access-requests list shows real data. The invite form either works end-to-end or is removed.

## Scope
- [ ] The pending-access-requests list queries the real `AccessRequests` collection for this resource and shows actual data (not a hardcoded string)
- [ ] A decision, documented in this issue's thread: build the real invite-by-email flow, or remove the disabled form
- [ ] If building the invite flow: a Payload-side mechanism for granting a user access to a resource with a given scope, wired to `inviteDeveloperByEmail`/`revokeDeveloperAccess`
- [ ] If removing the invite form: delete it and its TODO comment, leaving the tab either without it or with a clear "coming soon" note

## Out of Scope
- Building an email-sending mechanism - "invite" likely means granting access directly, not literally emailing anyone, unless maintainers want real email delivery (clarify in the issue thread before starting)
- Any other section of the resource management page

## Technical Notes
**Relevant files:**
[...]

Issue #270, opened 2026-08-30
[Bug] Developers see Approve/Deny buttons on their own submitted access requests


abubakr-itqan:
## Summary
On `/developer/requests` and `/dashboard/requests`, a developer viewing their own submitted access request sees Approve/Deny buttons on it, even though they're the applicant, not the resource owner.

## Background
Both pages feed `RequestCard` from `fetchMyRequests`, which queries `where[applicant][equals]=me` - i.e. the developer's own submitted requests. `RequestCard` shows Approve/Deny whenever the request is pending, with no ownership check in the UI.

The backend already rejects the actual mutation (`canUpdateAccessRequest` requires `resource.owner`), so this isn't a security hole - just a confusing UI that offers an action that will fail.

Surfaced while investigating #234.

## Scope
- [ ] `RequestCard` (or its caller) should not render Approve/Deny when the viewer is the applicant rather than the resource owner

## Acceptance Criteria
- [ ] A developer viewing their own pending access request no longer sees Approve/Deny buttons on it

## Difficulty
Easy

malakmuayad11:
Salam,

I read the issue description and would like to work on this issue. Please let me know if I can take it.

abubakr-itqan:
@malakmuayad11 shukran for the interest, assigning it to you now. It's a small, well-scoped one - the fix is just teaching RequestCard (or whoever renders it) to check that the viewer is the resource owner, not the applicant, before showing Approve/Deny. Take a look at CONTRIBUTING.md if you haven't already for setup and PR expectations.

abubakr-itqan:
@malakmuayad11 just checking in - it's been a couple weeks since this was assigned, no rush at all, just want to make sure it didn't slip off your radar. Let me know if you're still working on it or want a hand with anything.

malakmuayad11:
Hello @abubakr-itqan,

Thanks for following up. The issue has already been solved, and PR #274 has been merged to address this issue. If there is anything that needs modification, please let me know. I would also be happy to contribute to other issues if available.

abubakr-itqan:
Checked it myself - you're right, RequestCard already compares request.resource_owner_id against the logged-in user before showing Approve/Deny, from your own #274. Closing this out.

Since you're free now, feel free to grab anything from the open list: https://github.com/Itqan-community/RATQ/issues

abubakr-itqan:
Closed via Notion: status changed to **Done**.

PR #272, opened 2026-08-30
feat(access): let publishers view and revoke resource access

### LINKED ISSUE #234: [Feature] Let publishers view and revoke resource access
## Summary
A publisher has no way to see who currently has access to their resource, or to revoke it. The backend groundwork exists but is unused.

## Background
`src/modules/developer/infrastructure/access-api.ts` defines `inviteDeveloperByEmail` and `revokeDeveloperAccess`, with the file's own comment noting they're "currently unreferenced by any caller." `/developer/access` only manages API keys - it doesn't list who has access to a resource or provide any revoke action.

**Update after digging in further:** the `DELETE .../access/` endpoint this issue originally assumed exists does not - `access-api.ts` calls `/api/v1/developer/resources/{slug}/invite/` and `/api/v1/developer/resources/{slug}/access/`, which don't match anything in the current Payload backend (Payload's REST routes are collection-based, e.g. `/api/access-requests`, `/api/resources`). This looks like leftover scaffolding from an earlier API design that predates the actual Payload build. Also worth knowing: approving an access request today only flips that request's `status` to `approved` - no separate "access grant" record is created anywhere, so there's currently nothing for a revoke action to act on even if the endpoint existed.

## Scope
- [ ] Decide the data model for "who has access": most likely extending `AccessRequests` with a `revoked` status (or similar) rather than inventing a parallel grant record, since `approved` is already the closest thing to a grant today
- [ ] Add a real backend endpoint for listing/revoking access on a resource (the existing `access-api.ts` functions point at routes that don't exist and can't just be "wired up" as-is)
- [ ] Add a view (likely on the resource's dashboard/detail-management page) listing everyone with active access to a resource
- [ ] Wire a revoke action to the new endpoint
- [ ] Decide whether `inviteDeveloperByEmail` is in scope here too, or should be a separate follow-up (it's a distinct "grant access directly" flow vs. the request/approve flow that already works)

## Out of Scope
- The existing request/approve/deny flow (tracked separately in the "wire up built-but-unused interaction UI" issue) - this is specifically about *seeing and revoking* already-granted access

## Technical Notes
- `src/modules/developer/infrastructure/access-api.ts` - `inviteDeveloperByEmail`, `revokeDeveloperAccess` exist but point at a non-existent API namespace, not just "unused" - expect to rewrite this against the real backend
- `payload-backend/src/collections/AccessRequests.ts` - current source of truth for "approved" access; likely where the revoke state needs to live
- `src/app/developer/resources/[slug]/page.tsx` - the resource's own "إدارة الوصول" tab already has a scaffolded access-request list section (currently a hardcoded empty state) and a disabled invite form - this is a better home for the list/revoke UI than `/developer/access` (which is API-keys only, flat across resources). Also resolves #158, which asked for a build-or-remove decision on that same disabled invite form and was closed without the code actually changing.
- `src/modules/developer/domain/services/access-request-status.ts` - frontend mirror of the backend's status-finality guard; needs updating alongside the backend if the enum gains a `revoked` value
- `payload-backend/src/collections/APIKeys.ts` - `delete: isOwner` only allows the key's own holder to delete it, no admin/resource-owner branch. Revoking access needs to cascade to any API keys the developer already holds for that resource, or "takes effect immediately" in the AC isn't actually true - existing keys would keep working. Likely needs an `afterChange` hook on `AccessRequests` clearing the developer's keys via the local API, since `key_hash` isn't used for authentication anywhere yet (nothing to break) but the delete-access rule shouldn't be loosened
- Not in scope for this issue, but found while digging in: #270 (developers see Approve/Deny on their own submitted requests - UI-only, backend already rejects it)

## Acceptance Criteria
- [ ] A publisher can view a list of who has access to their resource
- [ ] A publisher can revoke a user's access, and it takes effect immediately (verified against the backend)

## Difficulty
Medium - this turned out to need real backend design work (a revoke-capable data model + a working endpoint), not just wiring existing pieces together. Flagging this before work starts so it's scoped accurately from the outset.

Jawad18750:
Closes #234.

Publishers can now see who holds access to a resource and take it back. Follows the decisions from the issue thread: option (a) for the data model, cascading to existing API keys, notifying the applicant, and building it into the resource's own access tab.

## The flow

Revoking access, Arabic

Arabic above, English below. Revoking takes two clicks because it cannot be undone.

Revoking access, English

## Backend

`revoked` joins the `status` enum. It is reachable only from `approved`, and everything the finality guard protected before is still protected: approved and denied still cannot be swapped, revoked is terminal, and a pending request has to be denied rather than revoked. That last one is a case I did not think about until I sat down to write the tests. The old guard only ran when the status was not `pending`, so a naive addition would have let a publisher jump a pending request straight to `revoked`, quietly skipping the review step and telling the applicant the wrong thing.

Revoking clears the applicant's API keys for that resource through the local API. This is the part that makes the acceptance criterion true rather than nominally true: `APIKeys.beforeValidate` only requires an approved request when a key is *created* and nothing re-checks afterwards, so a key issued while the request was approved would otherwise keep working forever. `APIKeys.delete` is scoped to the key's holder with no admin or resource-owner branch, so a publisher cannot clear them directly. As agreed, that rule is left alone and the cascade goes through the local API instead.

Both the key deletion and the notification run on the caller's transaction. Payload's Local API only joins it when handed `req`, and without that the delete commits on its own: a partial failure would leave the successfully-deleted keys permanently gone while the throw rolled back only the access-request row, so the request would revert to `approved` with some of the applicant's keys already destroyed. Checked against a real database, not reasoned about. With a failure raised after the delete, the keys go 2 to 0 without `req` and stay at 2 with it.

Revoking also emits the `access_revoked` notification to the applicant, matching how approved and denied already behave. The type, its DB enum value, its icon and its colour were all already in the codebase. Nothing had ever emitted it.

## Frontend

The resource's "إدارة الوصول" tab now lists everyone with access and offers a revoke on each row. The disabled invite form and the hardcoded "لا توجد طلبات وصول معلقة" are gone, and so is the dead `access-api.ts` chain. That turned out to be three files rather than one: two use-case wrappers imported it, so removing only the infrastructure file would have broken the build. That also settles the invite-form question #158 asked for and never got.

New copy goes through `ar.json` and `en.json`. Both files are pure additions, no reformatting, and the two locales have identical key sets.

One thing only browser testing caught. The confirm button first read "إلغاء الوصول" sitting right next to a Cancel button reading "إلغاء". Both start with the same word, so on an irreversible action the safe button could easily be read as the destructive one. The confirm button now reads "تأكيد الإلغاء" / "Confirm revoke", and there is a test pinning the two labels distinct.

## Testing

- backend: 169 to 186 tests
- frontend: 227 to 243 tests
- both typecheck clean; frontend lint unchanged from `main` at 26 problems

Test suites

I mutation-tested the new tests rather than trusting green: removing the API key cascade fails 3 of them, allowing `pending -> revoked` fails 2, making the first click revoke without confirming fails 4, and swallowing the revoke error fails 1. Each is caught by the test that should catch it.

For "verified against the backend" I ran the whole flow against a real Postgres with `overrideAccess: false`, so the collection access rules and hooks all applied exactly as an HTTP request would.

End-to-end backend verification

I also drove the real app in a browser against that database, in both locales: the list renders, the first click only asks for confirmation, cancelling leaves the grant alone, confirming removes the row and the list re-fetches from the backend, the empty state appears once nobody is left, and a non-API resource still shows the "API only" message. After the revokes the database had every grant `revoked`, zero API keys remaining, and one `access_revoked` notification per applicant. Logging in as one of those developers, the notification renders on their notifications page with the icon and colour that had been sitting unused in `developer/notifications/page.tsx`.

## Two things worth flagging

The generated `down()` fails if any request has actually been revoked by the time you roll back, with `invalid input value for enum`. I confirmed against a real database that the failure is atomic, so the enum, the rows and the migration record are all left untouched. I left it as generated and wrote a comment explaining why: the only way to make it succeed is to rewrite those rows first, and mapping revoked back to approved would silently hand back access a publisher deliberately removed. Rolling this back with live revoked data should be a deliberate human decision.

The migration adds 4 backend lint warnings for the unused `payload` and `req` params in the `up`/`down` signatures. That is what `payload migrate:create` emits, and all 13 existing migrations carry the same 4. I kept it consistent rather than being the one file that differs. No new lint errors in either workspace.

## Concurrency and small screens

Two publishers (or one publisher with two tabs open) can hit Revoke on the same grant at the same moment, so I fired five simultaneous `PATCH`es at one grant over HTTP. The status change and the API-key deletion are both idempotent: the grant ends up `revoked` once, the developer's keys are gone, and no other developer's keys are touched. Five `access_revoked` notifications are created though, one per request.

That last part is not something this branch introduces. I checked out `main` and raced five simultaneous approves at a single pending request, and the untouched hook produced five `access_approved` notifications the same way. Revoking inherits the existing behaviour rather than adding a new problem, so I have deliberately left it alone here instead of widening this PR into the approve/deny path. Happy to open it as its own issue if you want it fixed.

On a 390px viewport, both locales: no horizontal overflow in any state (`scrollWidth` equals `clientWidth` on the list and while confirming), each row stacks to name, date, then a full-width button, and the confirmation text wraps without clipping. The buttons are 36px tall, which is what `h-9` gives and exactly what `RequestCard`'s existing approve/deny buttons use, so this matches the rest of the dashboard rather than setting a new size.

## Not verified

The mobile checks were an emulated 390px viewport in Chrome, not a real handset, and I have not tried other browsers. The access list is fetched with `limit=100`, matching the existing api-keys and resources queries, so a resource with more than 100 active grants would silently show only the first hundred. That ceiling seems far off for now, but it is a real limit rather than something I tested past.

#270 is untouched, as agreed.


Jawad18750:
Thanks, this caught two real ones.

**API key delete errors:** correct, and worse than it looks. I checked, and `payload.delete` with a `where` filter returns `{docs, errors}` without throwing, so a failed deletion would have gone unnoticed and the applicant would still have been told their access was gone. I also tested what throwing from `afterChange` does, and it rolls the whole update back, so the hook now checks `errors` and throws. The revoke is atomic: either the status change, the key deletion and the notification all happen, or none of them do and the request stays `approved` for the publisher to retry. Better a failed revoke than a record that disagrees with reality.

**Empty state on a failed load:** also correct. I wrote the failing test first to be sure, and it did render "No one has access to this resource yet" on a rejected fetch, which is a very different claim from "we could not find out who has access". There is now an error state with a retry, and I verified it in a browser by intercepting the request to force a 500.

**The 100 cap:** I have not added paging, deliberately. Every list query in the app is capped at 100 the same way, and paginating only this one surface would be inconsistent. What I have changed is that the cap is no longer silent: the API returns the real total and the list says when it is showing a subset. Proper paging feels like its own issue covering the other lists too, happy to open it.

**Docstring coverage:** leaving this one. The codebase has around 580 line comments and 3 JSDoc blocks, and CONTRIBUTING does not ask for docstrings, so hitting 80% here would make these files the odd ones out. The new code is commented, just in the house style.

abubakr-itqan (changes requested):
Really thorough PR - the review discipline here (mutation-testing the new tests, verifying against a real Postgres with overrideAccess: false, checking the concurrency race against main before assuming this branch introduced it) is exactly what I want to see on something touching access control. Found one thing that needs a fix before this merges though.

`req.payload.delete(...)` and `req.payload.create(...)` in the `afterChange` hook don't pass `req` through. Payload's Local API only joins the caller's transaction when you explicitly hand it `req` - without it, it builds a fresh request object with no `transactionID`, so `payload.delete()` runs and commits in its own independent transaction before your hook even checks `errors`.

That breaks the atomicity guarantee the throw is there for: on a partial failure (some keys deleted, one fails), the successfully-deleted keys are already permanently committed by the time you throw, and the throw only rolls back the AccessRequests row itself - so the request reverts to `approved` while some of the applicant's keys for that resource are already gone. That's the "record and real access disagree" state your own comment says this code is meant to prevent, just reached from the opposite direction.

Fix should be one line at each call site - `req.payload.delete({ ..., req })` and `req.payload.create({ ..., req })` - so both join the parent update's transaction and a throw genuinely rolls everything back together. Worth a test that forces a partial delete failure (an error on one of several matched keys) to pin this, since the current tests stub `payload.delete` directly and wouldn't catch a transaction-boundary issue either way.

Everything else checked out clean on my end - tsc, lint (unchanged from main on frontend, exactly the expected +4 migration-boilerplate warnings on backend), the full test suites, the status-transition guard (all six transitions behave as described), the i18n parity between ar/en, and the AC against #234 with no scope creep. Once this one's sorted I'll get it merged.

Jawad18750:
Confirmed, and thank you for catching it. I tested it before fixing rather than taking it on trust, because I wanted to see the failure myself.

Two API keys on an approved request, with a failure raised after the delete. Without `req`, the keys went 2 to 0 while the row reverted to `approved`. With `req` passed, they came back, 2 to 2. So the delete really was committing independently and my throw was rolling back nothing but the access-request row. Then I re-ran it against the real hook by appending a throwing hook after it, and the keys, the notification and the status all rolled back together.

You are right that this is the same problem my comment claims to prevent, just from the other side. My previous fix was half a fix: it caught the reported failures but had no transaction to roll back into.

I have also passed `req` to the notification `create`. You did not raise it, but the same reasoning applies, and a notification telling someone their access was revoked would otherwise outlive a revoke that rolled back.

On the test: I have added the partial-failure case you asked for, one key deleted and one erroring, and assertions that `req` is threaded into both calls. I want to be straight about their limits though. They stub the Local API, so they pin the argument and not the transaction boundary; removing `req` from either call fails them, but they would not have caught the original bug on their own. The transaction behaviour is only verifiable against a real database, and that is where I checked it. A test that would genuinely catch it needs Postgres in CI, which is a bigger conversation than this PR.

Backend tests are 190 to 194. Lint and typecheck unchanged.

abubakr-itqan (approved):
Confirmed - req is now passed at both call sites (payload-backend/src/collections/AccessRequests.ts, the delete and the create), and I re-checked it independently: diffed the fix commit directly, ran tsc and lint clean (backend +4 warnings are still just the pre-existing migration boilerplate, 0 new errors), and ran the backend suite before and after your fix commit - 190 passing on the parent, 194 on this one, all four new tests real (not duplicates of the existing full-failure cases). Agree with your own read on the new tests too: they pin req at the argument level so nobody can silently remove it again, but the actual transaction-commit behavior only proved out against real Postgres, which is exactly what you did manually. That's the right call given there's no Postgres in CI yet.

Appreciate you testing the failure mode yourself (2 keys to 0, then 2 to 2) instead of taking my comment on faith - that's exactly the kind of verification I want to see on something touching access control. Merging.

PR #273, opened 2026-08-31
ci: serialize backend deploy jobs

### LINKED ISSUE #271: [Bug] Backend deploy races when main and beta push nearly simultaneously
## Summary

Merging a PR to `main` that also gets merged into `beta` (the standard main→beta sync) fires two independent "Deploy backend" workflow runs almost simultaneously - one triggered by the push to `main`, one by the resulting push to `beta`. Both scripts deploy to the same droplet checkout (which tracks `beta` regardless of which workflow fired), so whichever run finishes its Docker build first wins the container-name recreate, and the other fails with a name conflict.

## Evidence

Run `33299200645` (triggered by push to `main`, commit `e546329`) failed with:

```
Container ratq-payload-app-1 Recreate
Error response from daemon: Conflict. The container name "/82b8909dcbb0_ratq-payload-app-1" is already in
use by container "ce755e8666953cd0467a69c3f32521b32009282131a071d88dcceb6594beee7e". You have to remove
(or rename) that container to be able to reuse that name.
```

Timeline:
- Run `33299200645` (push to `main`) started its SSH deploy script at `07:26:52`.
- Run `33299214018` (push to `beta`, same merge event) started 21 seconds later at `07:27:13`.
- Both finished their Docker builds around `07:30:13`. The `beta` run won and recreated `ratq-payload-app-1` successfully; the `main` run hit the conflict a moment later.

Production was not affected - the `beta` run (the one that actually matters, since the droplet tracks `beta`) completed successfully. Only the `main` run showed red. This is at least the third time this exact symptom (container-name conflict) has shown up (see also PRs #242, #244), though those were caused by leftover orphaned containers rather than this race - worth checking whether this race has actually been the root cause of some of those too.

## Fix

Add a concurrency guard so two deploy runs can never hit the droplet at the same time. Options, roughly in order of preference:

- [ ] Add a `concurrency:` group to `.github/workflows/deploy-backend.yml` (e.g. `group: deploy-backend`, `cancel-in-progress: false` so the second run queues instead of racing) - simplest, handled entirely in GitHub Actions.
- [ ] Alternatively/additionally, add a `flock` around the deploy steps in the droplet-side deploy script so a second SSH session can't run concurrently even if two runners somehow overlap.
- [ ] Consider whether the `main`-triggered deploy is even needed, since the droplet's checkout only ever tracks `beta` - if not, gate the workflow to only run on `beta` pushes and let `main` merges deploy solely via the resulting `beta` push.

## Acceptance Criteria

- [ ] A main→beta merge no longer produces two simultaneous deploy runs racing on the same droplet target
- [ ] If both events still trigger a workflow, they queue rather than race (no container-name conflicts)

Walid-Khalfa:
Fixes #271

## Summary

Adds a concurrency guard to the backend deploy job to prevent simultaneous deployments to the same droplet.

## Changes

- Add job-level concurrency to the `deploy` job
- Use the same `deploy-backend` concurrency group for `main` and `beta`
- Set `cancel-in-progress: false` so an active deployment is not cancelled

## Why

Pushes to `main` and `beta` can occur nearly simultaneously during the normal main → beta sync. Both runs target the same droplet, which can cause a race during Docker container recreation.

Serializing the deploy job prevents this race while still allowing test jobs to run concurrently.

## Validation

- `git diff --check` passes
- Workflow structure reviewed
- Deployment triggers and scripts remain unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backend deployment reliability by ensuring deployments run one at a time.
  * Queued deployments are now preserved instead of being canceled.

### DIFF
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -36,6 +36,7 @@ jobs:
     concurrency:
       group: deploy-backend
       cancel-in-progress: false
+      queue: max
     runs-on: ubuntu-latest
     steps:
       - name: Deploy to droplet
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -33,6 +33,9 @@ jobs:
   deploy:
     needs: test
     if: github.event_name == 'push'
+    concurrency:
+      group: deploy-backend
+      cancel-in-progress: false
     runs-on: ubuntu-latest
     steps:
       - name: Deploy to droplet


Walid-Khalfa:
@abubakr-itqan , Jazak Allah khayran! The fix for #271 is ready for review. I kept the change minimal and followed the contribution guidelines. PR #273 is ready whenever you have a chance to take a look.

abubakr-itqan (approved):
Nice, small fix. Confirmed `queue: max` + `cancel-in-progress: false` is a valid, documented concurrency combo (the only restriction is that `queue: max` can't pair with `cancel-in-progress: true`), and since the `deploy` job already fires on push to both `main` and `beta`, sharing the `deploy-backend` group across both means the second push queues instead of racing the droplet - matches the first fix option from #271 exactly. Config-only change, no test needed per CONTRIBUTING. Merging.

Issue #269, opened 2026-08-30
[Bug] Store authentication session in HttpOnly cookie instead of localStorage


Ramahadam:
## Summary

RATQ currently stores the authenticated user's session JWT in browser `localStorage`.

Because `localStorage` is accessible to JavaScript running under the application's origin, an XSS vulnerability could read and exfiltrate the session token. A stolen valid token can then be replayed from another browser/session to authenticate as that user.

Move RATQ's authentication session from JavaScript-readable `localStorage` to Payload's HTTP-only cookie authentication.

Related: #229, which explicitly left broader localStorage token-storage hardening out of scope.

## Background

The current token storage is centralized in:

`src/shared/infrastructure/token-storage.ts`

which stores:

```text
ratq_access_token
ratq_refresh_token
```

in `localStorage`.

RATQ does not currently use separate access and refresh credentials. `payload-auth-repository.ts` documents that Payload issues a single JWT and that the same token is currently stored under both keys.

Payload already supports authentication using HTTP-only cookies, so this issue should use the existing Payload authentication mechanism rather than introduce a new refresh-token system.

## Scope

* [ ] Use Payload's HTTP-only cookie for the authenticated session
* [ ] Stop persisting the session JWT in `localStorage`
* [ ] Remove the duplicate `ratq_refresh_token` storage that currently contains the same JWT
* [ ] Update authenticated frontend requests to use cookie credentials
* [ ] Update login, logout, session restoration, and GitHub OAuth flows as needed
* [ ] Configure the cookie appropriately for production (`HttpOnly`, `Secure`, and suitable `SameSite` behavior)
* [ ] Add regression coverage for the changed authentication flow

## Technical Notes

Because the frontend and Payload API are served separately, the implementation may also require updating credential/CORS/CSRF configuration so the browser can send the authentication cookie to Payload.

The cookie should remain inaccessible to client-side JavaScript.

Moving the token to an HttpOnly cookie reduces the impact of XSS by preventing JavaScript from directly reading and exfiltrating the session credential. It does not replace normal XSS protection.

## Acceptance Criteria

* [ ] No authentication/session JWT is stored in `localStorage`
* [ ] The session credential cannot be read through client-side JavaScript
* [ ] Login works using the cookie-based session
* [ ] Authentication survives page refresh
* [ ] Authenticated API requests continue to work
* [ ] Logout clears/invalidates the authenticated session
* [ ] Session expiry behavior continues to work
[...]

PR #281, opened 2026-09-03
Feat/adding http only cookie

MahmoudSheemy127:
Problem: Replacing Authorization Bearer token with HttpOnly Cookie to avoid XSS attacks, Related to [issue](https://github.com/Itqan-community/RATQ/issues/269)

Overview about the implementation done for each file through the PR:

- **payload-backend/src/collections/Users.ts**: Updated by Initializing and defining HttpOnly Cookie parameters including (secure parameter, sameSite parameter to avoid CSRF attacks, Expiration & others.)
- **payload-backend/src/app/oauth/github/exchange/route.ts**: Adding new cookie configuration for the github OAuth route, to return HttpOnly Cookie instead of the signed JWT token.
- **src/modules/auth/infrastructure/payload-auth-repository.ts**: Removing the Authorization Bearer token for each fetch call and replacing it with credentials: 'include'. Adding new async function **fetchUserDetails** To fetch user details utilizing the browser cookie.
- **src/hooks/useAuth.tsx**: Doing refinement to apply with the new Cookie migration including adding an extra hook **getUserData** which would get executed with the pre page render to authenticate and fetch user details utilizing the browser cookie, Instead of checking the local storage for the token or user details stored in the browser. Also doing enhancements across the pre-existed hooks such as **login**, **logout** and **register** to avoid utilizing the local storage tokens. Also updating hook **loginWithCode** for the OAuth route to apply with the new Cookie migration 
- **src/modules/resources/infrastructure/reports-api.ts**: Replacing Bearer token with HttpOnly Cookie through the fetch calls
- **src/modules/resources/infrastructure/comments-api.ts**: Replacing Bearer token with HttpOnly Cookie through the fetch calls
- **src/modules/resources/infrastructure/access-requests-api.ts**: Replacing Bearer token with HttpOnly Cookie through the fetch calls
- **src/modules/developer/infrastructure/resources-api.ts**: Replacing Bearer token with HttpOnly Cookie through the fetch calls
- **src/shared/infrastructure/payload-error.ts**: Updating Session check methods to apply with the new Cookie Migration
- Updating test scripts to apply with the new Cookie migration

Summary:
- [x] All fetch calls using HttpOnly Cookie
- [x] tokens and refresh tokens are no longer being stored in local storage
- [x] Github OAuth flow works seamlessly with the new Cookie migration strategy 
- [x] Front end test coverage
- [x] Backend test coverage

## Summary by CodeRabbit

* **New Features**
  * Authentication now uses secure browser sessions for sign-in, registration, OAuth, logout, and session restoration.
  * Added automatic session-expiration detection and sign-out handling.
  * API requests for resources, notifications, developer tools, and access requests now use the active browser session.
* **Security**
  * Added login attempt limits and temporary account lockouts.
  * Session cookies now use environment-appropriate security settings.
* **Configuration**
  * Updated backend runtime port configuration.
  * Certificate files are excluded from version control.


abubakr-itqan:
Shukran for taking this on, this is genuinely one of the more security-sensitive changes we've had, and you got the part most likely to go silently wrong right - the `payload-token` cookie name matches Payload's real default, which I checked against the framework docs directly, since a wrong guess there would have broken OAuth login while every existing test kept passing.

A few things need fixing before this is ready:

- `useAuth.tsx:100` has a new lint error - `getUserData` is used inside a `useEffect` before it's declared further down in the same function. `npm run lint` doesn't run in CI here so it slipped through, but it's a real one.
- The PR checklist says backend test coverage is done, but no backend test files changed - this touches the cookie config and login lockout, which CONTRIBUTING.md calls out as needing tests since it's auth-path logic.
- `getUserFromStorage()` in `useAuth.tsx` and `getAccessToken`/`setAuthTokens`/`clearAuth` in `token-storage.ts` are commented out rather than deleted - CONTRIBUTING.md asks for the old helper to be removed once nothing references it.
- In the register flow, `register()` already calls `login()` internally and that now sets the cookie too, so the extra `loginUseCase()` call added in `useAuth.tsx`'s register handler ends up logging in twice on every signup.

Also curious about the Dockerfile/`.env.example`/`.gitignore` changes (port 4000, `certificates`) - not obviously related to the cookie migration, a quick note on why would help review.

Nothing here is a design problem, the approach is right - just needs the loose ends tied up. Ping me once it's updated and I'll take another look.

abubakr-itqan (changes requested):
Reviewed this locally - really solid work on the redemption/CSRF-token logic and the OAuth flow, and the test coverage update is thorough. Two things I'd like fixed before merging though:

1. `src/hooks/useAuth.tsx:100` calls `getUserData()` inside the mount effect before its `useCallback` declaration later in the component - it happens to work since effects run after render, but it's a genuine new lint error (`Cannot access variable before it is declared`), plus it drags in an unused eslint-disable and a missing-dependency warning. Moving the `getUserData` callback above the effect (or inlining the fetch) clears all three.
2. In both `payload-backend/src/collections/Users.ts` and `payload-backend/src/app/oauth/github/exchange/route.ts`, the cookie's `domain` is set to `.itqan.dev` in production. That scopes the session cookie to every itqan.dev subdomain, not just RATQ's (`ratq.itqan.dev` / `beta.ratq.itqan.dev` / `admin.ratq.itqan.dev` per the CORS allowlist in `payload.config.ts`). Since browsers treat same-eTLD+1 subdomains as "same-site," `SameSite=Lax` won't stop another itqan.dev app from riding this cookie to RATQ's API - could you scope it to `.ratq.itqan.dev` instead?

Also noticed `token-storage.ts` and `useAuth.tsx` have some now-dead code left commented out rather than removed (`getAccessToken`/`setAuthTokens`/`clearAuth`, the old `getUserFromStorage`) - worth deleting since nothing calls them anymore.

Happy to take another look once these are in.

MahmoudSheemy127:
Dear @abubakr-itqan  Al Salamu Alaykum

Thanks for your precious feedback. I have done the remaining enhancement and refactors reported, which include the following:

1- **src/hooks/useAuth.tsx:100**: Updated the positioning of the getUserData callback function to comply with the es-lint check and to follow the best coding practice
2-  **payload-backend/src/collections/Users.ts**: Domain field is updated to include all of .**ratq*** sub-domains, to safely apply valid and safe **sameSite** setting only on the **ratq** sub-domains
3- Cleaned old code regarding token storage and fetching from script files **token-storage.ts** and **useAuth.tsx**
4- Test coverage: You can find initially the PR included some test coverage regarding the new cookie setting which you can find that it is covered in the following test files **useAuth.test.ts**x, **resourceAccessApi.test.ts**, **payloadError.test.ts** & **AuthErrorNavigation.test.tsx**.

Waiting for your feedback inshallah on these set of changes, and let me know if there is anything that would require and extra checkup.

Thank you.

abubakr-itqan (changes requested):
Nice, this is much closer. `useAuth.tsx` and `token-storage.ts` are clean now, lint's back to baseline (8 errors, matches main, none new), tsc's clean, and the full suite still passes 264/264.

One thing left though: the cookie domain fix only landed in `payload-backend/src/collections/Users.ts`. `payload-backend/src/app/oauth/github/exchange/route.ts` sets its own `payload-token` cookie separately (that's the one issued after the GitHub OAuth code exchange) and it's still got `domain: process.env.NODE_ENV === 'production' ? '.itqan.dev' : undefined` on line 129 - same overly-broad scoping issue, just in the other spot that sets this cookie. Could you mirror the same `.ratq.itqan.dev` fix there too? Once that's in I think this is good to go.

abubakr-itqan:
@MahmoudSheemy127 just following up on the last review - the fix for the OAuth exchange route's cookie domain is the only thing left blocking this. Let me know once it's in and I'll take another look.

MahmoudSheemy127:
@abubakr-itqan Salam

Github OAtuh Cookie domain has been updated to include the **.ratq.itqan.dev** subdomain

abubakr-itqan (approved):
Confirmed both cookie domains are now scoped to .ratq.itqan.dev in Users.ts and the OAuth exchange route. Re-ran everything on the final branch: tsc clean, lint back to baseline (8 errors matching main exactly, none new), full suite passing 264/264. This addresses #269 properly. Merging.

PR #288, opened 2026-09-12
feat: normalize Arabic text for search matching

### LINKED ISSUE #287: Normalize Arabic text in the search
What

Normalize Arabic text before matching search queries against resource names/descriptions, so that variant letter forms are treated as equivalent. Specifically:

Alef variants (أ, إ, آ, ٱ) normalized to bare alef (ا)
Alef maksura (ى) normalized to yaa (ي)
Tashkeel/diacritics (fatha, kasra, damma, sukun, shadda, etc.) stripped before comparison
Taa marbuta (ة) vs haa (ه) — open question, see below

This would live entirely in the frontend/data-layer: a new normalizeArabic() utility in src/shared/utils/utils.ts, applied to both the search query and the fields being matched (name, description) in each of the three resource sources (cms.ts, payload.ts, ratq-native.ts).

Why

Right now, search does a plain .toLowerCase().includes() match with no Arabic-specific normalization. A search for "الدورى" (with alef maksura) won't match a resource titled "الدوري" (with yaa), even though most Arabic speakers consider these the same word — they differ only in typing convention/regional spelling habits, not meaning. This creates false "no results" for legitimate searches and makes discovery feel unreliable for Arabic-first users, who are a core audience for a Quranic-data marketplace.

I traced the full search path (form → SWR hook → use-case → API route → aggregator → per-source filter) and confirmed:

Search is entirely in-memory in this repo; neither the CMS nor Payload backend receives or needs to know about the search query, so this is fully fixable here with no backend changes.
All three sources duplicate the same filter block independently, matching the existing per-source-owns-its-filtering convention (the aggregator only sorts/paginates).

Open question for maintainer input: the taa marbuta (ة) → haa (ه) normalization is more contentious than the others — it would help a search for "رحمه" match "رحمة", but risks false positives where a trailing ه is intentionally distinct from ة. I want guidance on whether to include this mapping, make it configurable, or leave it out for v1.

Note: I've already worked out a solution for this (new utility + updated filter logic in all three sources, with unit and integration tests) and will link a PR against this issue for review.

Scope note

This is for RATQ platform features (catalog, filtering, discovery, developer dashboard, etc.). For new resources to list, please use the connected data sources (e.g., the CMS), not manual submissions here.

Acceptance criteria
- [ ] A maintainer confirms this fits RATQ's roadmap before implementation starts
- [ ] Maintainer input received on whether to normalize taa marbuta (ة) ↔ haa (ه)
- [ ] normalizeArabic() utility added with unit tests covering alef variants, alef maksura/yaa, tashkeel stripping, and mixed Arabic/non-Arabic input
- [ ] Search filter logic updated consistently across cms.ts, payload.ts, and ratq-native.ts
- [ ] At least one integration test confirming a normalized search (e.g. "الدورى" ↔ "الدوري") returns matching results

Abdalluh28:
Closes #287

## What
Adds `normalizeArabic()` to `utils.ts` and applies it in `cms.ts`, `payload.ts`, and `ratq-native.ts` so Arabic letter-form variants (alef forms, alef maksura/yaa, tashkeel) match as equivalent in search.

Taa marbuta ↔ haa normalization left out pending maintainer input per the open question in #287.

## Why
Search previously did a plain `.toLowerCase().includes()` with no Arabic normalization, so e.g. "الدورى" wouldn't match "الدوري". Fully frontend — search is in-memory; backends never see the query.

## Testing
- Unit tests for `normalizeArabic()` (alef variants, alef maksura/yaa, tashkeel, mixed input, empty string)
- Integration test confirming "الدورى" ↔ "الدوري" match in `ratq-native.ts`
- `npm run lint` and `npm test` pass, no new lint errors

## Summary by CodeRabbit

* **New Features**
  * Arabic resource searches now normalize common spelling variations, diacritics, and letter forms for more consistent matching.
  * Search remains case-insensitive for English text while preserving support for Arabic and Arabic-Indic numerals.

* **Tests**
  * Added coverage for Arabic text normalization and resource search behavior, including matching accuracy and no-search result handling.


abubakr-itqan (approved):
Reviewed this against #287 - tsc clean, lint at baseline (26 problems, matches main exactly, nothing new), and the test suite is thorough (298/298 passing, including the new normalizeArabic and ratq-native search coverage). The alef-variant, alef-maqsura, and tashkeel normalization all match the issue exactly.

One thing worth flagging for the record: the PR description says taa marbuta (ة) to haa (ه) normalization was "left out pending maintainer input," but it's actually implemented - `utils.ts` has `.replace(/ة/g, 'ه')` and both test files assert on it. Went ahead and made the call: keeping it in is fine, "رحمة" matching "رحمه" is a reasonable win for search recall. Could you fix the PR description so it matches what's actually shipping though - for next time, it's worth double-checking the description against the diff before opening, especially on an open design question like this one.

Merging as-is.

Abdalluh28:
Ah good catch, to clarify, the intent was that taa marbuta/haa was an open question I wanted your input on before it got merged, not something I forgot to add. Should've worded the description more clearly. Glad it works for you as-is. Thanks for reviewing!

Also, this is my first open source contribution, so I really appreciate the thorough review and the guidance along the way 🙏

abubakr-itqan:
That makes total sense, jazak Allah khayran for clarifying! And honestly, that's a great instinct to have going forward - flag the open question in the PR itself, not just the issue - it'll save the next reviewer a double-take like I had.

And genuinely, shukran for such a solid first contribution. The normalization logic was clean, the test coverage was thorough (edge cases like superscript alef and Quranic pause marks included, which most people wouldn't even think to test), and you traced the whole search path before touching anything instead of guessing. That's a strong start, welcome to RATQ, and I hope this is the first of many in-shaa-Allah 🙌

PR #290, opened 2026-09-13
fix: redirect authenticated users away from register page

### LINKED ISSUE #289: [Bug] The authenticated users can go to the registration page
## What's wrong

A logged-in user can navigate to `/register` and see/use the registration 
form. Expected behavior: they should be redirected away (e.g. to 
`/dashboard`), same as what already happens on `/login`.

## Where

https://beta.ratq.itqan.dev/register
`src/app/register/page.tsx`

## Steps to reproduce

1. Log in
2. Navigate to `/register`
3. The registration form loads normally instead of redirecting, and you can register with another email

For comparison, `src/app/login/page.tsx` correctly redirects an 
authenticated user to `/dashboard` via a `useAuth()` check — `RegisterPage` 
has no equivalent guard; it only calls `clearError()` on mount.

Note: I've already implemented the fix (mirroring `LoginPage`'s guard) and 
will link a PR against this issue immediately.

## Acceptance criteria

- [ ] The bug is fixed

Abdalluh28:
Closes #289

## What
Adds the same authenticated-user redirect guard from `LoginPage` to `RegisterPage`. If `user` is present (and `loading` is false), redirects to `/dashboard` instead of rendering the registration form, with the same spinner shown while the auth check resolves.

## Why
`LoginPage` already guards against this; `RegisterPage` had no equivalent check, only calling `clearError()` on mount. This let a logged-in user land on and use the registration form.

## Testing
- `npm run lint` and `npm test` pass; no new failures

## Summary by CodeRabbit

* **New Features**
  * Added authentication-aware registration flow.
  * Authenticated users are redirected to the dashboard.
  * A loading indicator is shown while authentication status is being determined.
  * Registration form is hidden for authenticated users.


abubakr-itqan (approved):
Checked out and ran it - tsc clean, lint at baseline (28 problems matching main exactly, nothing new), and the full suite passes 299/299 including your new RegisterPage tests. Also compared it against LoginPage's existing guard directly and it's a faithful mirror of the same pattern, not just a claimed match. Nice touch fixing the AuthErrorNavigation test to track the context-level error instead of the now-unmounted form's local state - that could've easily been left flaky. Merging.

PR #291, opened 2026-09-13
fix: changed the logo to be with the right spelling

### LINKED ISSUE #285: Update logo asset for the corrected Arabic app name
## Where

`public/images/logo.png`

## What needs fixing

The logo asset still needs to be updated to reflect the corrected Arabic app name spelling: `رَتْق`.

## Acceptance criteria

- [ ] The logo uses the corrected Arabic spelling `رَتْق`
- [ ] The updated asset remains clear and visually consistent across the app
- [ ] Existing logo usages continue to render correctly

Abdalla2200:
Closes: #285 

### **What changed**

- Updated `public/images/logo.png` with the corrected Arabic spelling رَتْق.
- Kept the existing logo design and appearance unchanged.


abubakr-itqan (changes requested):
Checked the actual pixels, not just the diff stat (CodeRabbit's review was skipped on this file due to its default PNG path filter, so I looked closely myself). The spelling correction itself is right - I zoomed into the diacritic and confirmed the old logo has a fatha stroke over the taa (رَتَق) while the new one has a proper sukun circle (رَتْق), exactly what #285 asked for.

But the file itself needs work before this can merge: it went from 1.3KB to 226KB, a 165x increase, and it's now a 966x700 3D-embossed render instead of the old flat 86x59 wordmark. This logo is used at roughly 30px across the header, footer, sidebar, and every auth page (login/register/reset/verify) - that's a real page-weight hit on every single page load for an asset nobody will ever see at more than icon size.

Could you re-export it at something like 128x128 or 256x256 (plenty for retina at 30-34px display size) and ideally flatten the 3D bevel/shadow effect back to the site's existing flat, minimal look - that'll also make the PNG compress down much smaller on its own. Happy to take another look once that's in.

Abdalla2200:
@abubakr-itqan I updated the logo based on your feedback. The new file is 86×62 and 1.13 KB, with the correct **رَتْق** spelling and solid black color. Its display size is also almost the same as the original.

Could you please take another look?

abubakr-itqan (approved):
Checked the file directly - 1160 bytes, 86x62px, basically the same footprint as the original (86x59). Zoomed into the diacritic again too, the sukun circle over the taa is still there so the spelling fix held up through the re-export. This is good to go. Merging.

PR #306, opened 2026-09-16
redesign filter panel with multi-select CC license filter

### LINKED ISSUE #298: [Design] Redesign the resources filter panel and fix the license filter
> **Difficulty:** Medium-High　**Good First Issue?:** No - touches the URL param and all three data sources, not just this component　**Estimated effort:** About 2 days (visual restyle plus multi-select logic across the app)

## Summary
The filters on the left of the resources page are plain black-and-white buttons with no real design. The new design wants a proper card with a title, a clear-all button, and colorful tags for each resource type.

Design reference: full filter panel
Design reference: license checkboxes group

## Background
Today's filter panel is very basic - no card background, no colors, and the "clear all" option is just a plain text link instead of a proper button with an icon.

Looking closely at the license filters in the design, they turn out to be quite different from what we have. The design shows checkboxes for content licenses (like Creative Commons - CC0, CC BY, etc.), letting you pick more than one at a time, each shown with a small colored badge.

Today's license list already includes a few Creative Commons options (like CC-BY-4.0) alongside software licenses (like MIT), so it isn't a completely different list - but it's missing some the design shows (like plain CC0), and more importantly it only lets you pick one license at a time, where the design's checkboxes let you pick several. Decided: move fully to the design's Creative-Commons-only list, with multiple selections allowed.

One catch: the design's checkboxes show short display labels (CC0, CC BY, CC BY-SA...), but the actual stored license values are the longer, versioned strings (CC-BY-4.0, CC-BY-SA-4.0, etc.), and today's filtering is an exact string match. Picking "CC BY" needs to match every stored value that belongs to that label (e.g. CC-BY-4.0 and any other CC-BY version), not just look for a literal "CC BY" that will never exist in the data. This needs a small label-to-stored-values mapping, not a one-to-one rename. It also needs a decision on what happens to resources whose license isn't one of the design's rows (MIT, Apache, GPL, BSD, custom - plus at least one Creative Commons variant, CC-BY-NC-4.0, that doesn't have a row in the design either) once they're no longer offered as filter choices - they should simply become unfilterable-by-license, not disappear from the catalog.

This also touches more than just this component - the URL parameter, the request layer, and the filtering logic in all three data sources need to move from a single value to a list of values.

Also worth flagging: one row in the design looks like a mistake - two different licenses are shown with the same badge label by accident. Point this out to whoever owns the design rather than copying the mistake into the real product.

The colors for each resource type (library, dataset, etc.) are already defined elsewhere in the code, so this task should reuse them rather than making up new colors.

## Scope
- [ ] Pull the existing resource-type colors out into one shared place, and use that in the filter panel too, instead of each place defining its own copy or a different palette.
- [ ] Give the filter panel a proper card look: border, rounded corners, spacing, matching the design.
- [ ] Add the subtitle text and a proper clear-all button with an icon.
- [ ] Show resource types as colorful rounded tags instead of plain black-and-white buttons (visual only - which type is selected still works exactly as it does today).
- [ ] Move the license filter to the design's Creative-Commons-only checkboxes, with a mapping from each display label to the stored license value(s) it should match.
- [ ] Change the license filter from one-at-a-time to multiple selections - update the URL parameter (as repeated values, e.g. `?license=a&license=b`, matching how the rest of the app already reads URL parameters) and the filtering logic in all three data sources.
- [ ] Decide and document what a non-Creative-Commons resource shows once its license is no longer a filter option (it should just be unfilterable by license, not hidden from the catalog).
- [ ] Flag the licensing design mistake mentioned above to whoever owns the design.
- [ ] Add the new text in both Arabic and English.

## Out of Scope
- Adding a publisher filter - tracked in a separate issue.
- Changing the resource-type filter's behavior (which type is selected, how it filters) - only its look changes here.
- The search bar and results-count issues.

## Technical Notes
- The filter panel code is `FilterPanel.tsx`.
- The resource-type colors currently exist in two separate copies (`ResourceDetailClient.tsx` and `ResourceCard.tsx`), plus a third, different-looking badge used elsewhere - reconciling those into one shared source is part of this issue, not a prerequisite for it.
- Today's license filtering is a single value threaded through the catalog page, the URL-building helper, the API route, and each of the three data sources' own equality check - all of those need to move from one value to a list.
- There's an existing test around combined type+license filtering that currently assumes a single license value; it'll need updating alongside the multi-select change.

## Acceptance Criteria
- [ ] The license filter shows the Creative-Commons-only checkboxes from the design, and picking one correctly matches every stored license value that belongs under that label.
- [ ] More than one license can be picked and cleared together via clear-all.
- [ ] The filter panel looks like a proper card, matching the design.
- [ ] Each resource-type tag uses the same color as that type's chip on the resource detail page.
- [ ] Picking a resource type still filters results exactly as it does today - only its appearance changed.
- [ ] Looks correct in both Arabic and English.

Abdalluh28:
Closes #298 

## Summary

Redesigns the Resource Library filter panel and converts the Creative Commons
license filter from a single-select value into a multi-select checklist.

### Filter-panel redesign

- Replaced the plain filter layout with a bordered, rounded card panel.
- Added filter subtitle and an always-visible "Clear all" action with a trash icon.
- Reworked resource-type filters into colorful, rounded tags with icons.
- Extracted shared resource-type color/icon definitions and reused them across
  `FilterPanel`, `ResourceCard`, and `ResourceDetailClient`.

### Multi-select CC license filtering

- Replaced the old single-value license buttons with six Creative Commons
  checkbox rows: CC0, CC BY, CC BY-SA, CC BY-ND, CC BY-NC-SA, CC BY-NC-ND.
- A checkbox selects every stored license alias associated with that CC family.
- Multiple CC families can be selected together; toggling one row preserves
  selections from other rows.
- "Clear all" clears both type and license selections and resets pagination.

### URL and request-chain support

- Changed the license query format from a single value to repeated params:
  `?license=value-a&license=value-b`
- Updated `FilterPanel`, `CatalogContent`, `useResources`, the request URL
  builder, the API route, `ResourceListParams`, and the aggregator to support
  `license?: string[]`.
- The API route reads repeated values with `searchParams.getAll('license')`.

### Resource-source filtering

- Added `shared/utils/license-filter.ts` as the single source of truth for
  the CC license mapping (`CC_LICENSE_ROWS`) and the filtering rule
  (`matchesLicenseFilter`, `TRACKED_LICENSE_VALUES`), consumed by all three
  data sources (RATQ-native, CMS, Payload) and re-exported from `FilterPanel`
  for the UI.
- License filtering uses OR logic across selected values.
- Resources with licenses not represented by the CC checklist remain visible
  and are never filtered out — this includes MIT, Apache, GPL, BSD, custom
  licenses, and CC-BY-NC-4.0 (intentionally has no checkbox row).

### Tests

- Replaced the outdated single-select `FilterPanel` tests with multi-select
  coverage (rendering, combined type+license filtering, add/remove per row,
  page-reset, Clear-all always-visible).
- Added `license-filter.test.ts` covering tracked vs. untracked licenses,
  the CC-BY-NC-4.0 no-row case, and consistency between `TRACKED_LICENSE_VALUES`
  and `CC_LICENSE_ROWS`.

## Summary by CodeRabbit

- **New Features**
  - Added multi-select Creative Commons license filtering.
  - License selections can combine multiple values and persist in the URL.
  - Added grouped license options with descriptions, badges, and localized English and Arabic labels.
  - Redesigned resource filters with colored type buttons, icons, responsive layout, and an always-visible “Clear all” button.
  - Added consistent resource-type colors and icons across cards, details, and filters.

- **Bug Fixes**
  - Improved license filtering consistency across resource sources.
  - Fixed pagination reset behavior when filters change.


abubakr-itqan (approved):
Went through this properly - tsc clean, lint at baseline (28 problems, matches main exactly, nothing new), and 321/321 tests pass (22 new ones covering every branch of the multi-select toggle logic, the CC0-1.0/CC0 alias variants, and the CC-BY-NC-4.0 pass-through case explicitly). Checked matchesLicenseFilter and TRACKED_LICENSE_VALUES directly against the issue's AC - untracked licenses (MIT, Apache, custom, CC-BY-NC-4.0) really do stay visible and unfilterable, exactly as decided. Good call pulling the color/icon maps into shared constants too, that was duplicated three ways before.

Two tiny things, not worth blocking on: src/modules/resources/application/use-cases/list-resources.ts is now fully unused since useResources.ts calls fetchResources directly - fine to leave for a cleanup pass later. And FilterPanel.tsx is missing its trailing newline.

Merging.

abubakr-itqan:
CodeRabbit's post-merge review caught something real here - confirmed it live on prod: `GET /api/resources?license=` (an empty value) drops the catalog from 118 resources to 20, hiding every tracked CC-licensed resource instead of falling back to no filter. Not reachable through normal clicks in FilterPanel today, but a malformed/shared URL can trigger it. Opened #307 to track the fix - it's a small one, contained to matchesLicenseFilter.

PR #308, opened 2026-09-16
fix: ignore unrecognised license values in matchesLicenseFilter

### LINKED ISSUE #307: [Bug] Empty or unsupported license param hides all tracked CC resources
## Summary

A URL with an empty or unsupported `license` query param hides every tracked Creative Commons resource from the catalog instead of being ignored, since PR #306.

## Where

`src/shared/utils/license-filter.ts` (`matchesLicenseFilter`)

## Steps to reproduce

Hit the resources API directly with an empty license value:

```
GET /api/resources?license=
```

Confirmed live on prod right now: with `license=` in the URL, the catalog drops from 118 total resources to 20 - every one of the 98 resources with a tracked CC license (89 CC-BY-NC-ND, 7 CC-BY-ND, 1 CC0 1.0, 1 CC-BY-SA 4.0) disappears, leaving only the 20 with an untracked license (Unspecified, MIT, GPL-3.0, Waqf Public License). A garbage/unrecognized license value (e.g. `?license=not-a-real-license`) triggers the exact same behavior.

## Background

`matchesLicenseFilter` treats any non-empty `selected` array as an active filter:

```ts
export function matchesLicenseFilter(resourceLicense: string, selected: string[] | undefined): boolean {
  if (!selected || selected.length === 0) return true;
  if (!TRACKED_LICENSE_VALUES.has(resourceLicense)) return true;
  return selected.includes(resourceLicense);
}
```

The problem is `selected` isn't validated - it can contain values that aren't in `TRACKED_LICENSE_VALUES` at all (an empty string, a typo, a stale/removed license value, anything from a hand-edited or shared URL). None of those values will ever equal a real resource's license, so every tracked-license resource fails the `selected.includes(resourceLicense)` check and gets hidden, while untracked-license resources still pass through via the line above. The result reads as "the catalog is nearly empty" rather than "your license filter didn't match anything real."

This can't happen through normal clicks in `FilterPanel` today (it only ever toggles real values from `CC_LICENSE_ROWS`), but it's reachable by anyone who lands on a hand-typed, bookmarked, or shared URL with a malformed `license` param.

## Scope

- [ ] `matchesLicenseFilter` (or its caller) should ignore any `selected` value that isn't in `TRACKED_LICENSE_VALUES` before deciding whether a filter is "active" - an empty or fully-unrecognized selection should behave the same as no filter at all (show everything), not "show nothing tracked."
- [ ] A selection that mixes valid and invalid values (e.g. `?license=CC-BY-4.0&license=`) should still filter correctly on the valid value(s) and simply drop the invalid one.

## Out of Scope

- Changing what counts as a "tracked" license - that's #298's decided scope, unrelated to this bug.
- Any change to `FilterPanel`'s UI - it never produces an invalid value today, this is purely about making the underlying filter function defensive against a malformed URL.

## Technical Notes

- The fix likely belongs in `matchesLicenseFilter` itself in `src/shared/utils/license-filter.ts`, since all three data sources (`ratq-native.ts`, `cms.ts`, `payload.ts`) call it directly and would all be fixed at once.
- `license-filter.test.ts` already has good coverage to extend - add cases for an empty-string value and an unrecognized value, both alone and mixed with a valid value.

## Acceptance Criteria

- [ ] `GET /api/resources?license=` returns the full, unfiltered catalog (same result as no `license` param at all).
- [ ] `GET /api/resources?license=not-a-real-license` also returns the full, unfiltered catalog.
- [ ] `GET /api/resources?license=CC-BY-4.0&license=` (one valid value, one empty) returns only resources matching `CC-BY-4.0`, not an empty result.
- [ ] Existing valid-selection filtering behavior (from #298) is unchanged.

Abdalluh28:
Fixes a bug where invalid or empty license URL parameters (e.g., ?license=invalid) incorrectly hid all tracked resources from the catalog. While unreachable via the UI, this occurred when users visited modified or malformed URLs.

Changes

Bugfix: Strips unrecognized values from the selected array by validating against TRACKED_LICENSE_VALUES before applying the filter. If the array is empty, it safely defaults to "no filter".

Testing: Added 6 new test cases covering empty strings, unrecognised values, and mixed valid/invalid inputs.

Closes: #307

## Summary by CodeRabbit

* **Bug Fixes**
  * License filtering now ignores empty, unrecognized, or outdated filter values.
  * Valid license filters continue to match tracked licenses while allowing untracked licenses through.
  * Invalid-only filter selections no longer incorrectly exclude resources.


abubakr-itqan (approved):
Fast turnaround, jazak Allah khayran. Checked it against exactly what #307 needed - validSelected filters selected down to TRACKED_LICENSE_VALUES before checking length, so an empty or garbage-only selection now falls back to no-filter instead of hiding everything, and a mix of valid+invalid correctly filters on just the valid part. tsc clean, lint at baseline (28 problems, matches main), 327/327 tests pass including your 6 new cases. Merging.

PR #311, opened 2026-09-20
Feat/rebuild the homepage announcements

### LINKED ISSUE #304: [Design] Rebuild the homepage announcements banner to the Figma
> **Difficulty:** Medium　**Good First Issue?:** No - visual rebuild plus reworking rotation without visible controls　**Estimated effort:** 1 day

## Summary
The design's homepage announcement banner is a single black bar, not the card-with-dots carousel we have today. This is a full visual rebuild, and includes reworking the rotation behavior to have no visible controls.

Design reference: homepage announcements banner

## Background
The real design for this: it's one plain black bar across the page, with a section label on one side, the announcement text in the middle (starting with a bold "New:"), and a button on the other side. There are no dots, arrows, or counters visible.

This is quite different from what exists today: a padded white card with a colored side bar, a badge, a timestamp, and next/previous buttons with dots for flipping between several announcements.

Decided: keep rotating through multiple announcements automatically, just without any visible arrows, dots, or counter - the timer-based auto-rotate stays. Keyboard navigation needs a small fix while we're in here: today it listens for arrow keys on the whole page, which only sort-of works because the visible arrow buttons hint that arrow keys do something. Once those buttons disappear, that page-wide listening would hijack arrow keys with no visible reason why - so it should either be scoped to just the banner (only react when the banner itself is focused) or dropped, now that there's no visible control it's paired with.

Also decided: drop the four announcement types' distinct colors and icons entirely - one plain black bar for every type. This is a styling change only: one of the announcement types today also changes where its button links to (to a specific resource page instead of a general link) - that routing behavior stays exactly as it is, only the color/icon styling goes away.

## Scope
- [ ] Rebuild the banner as a single black bar matching the design: label on one side, announcement text in the middle, button on the other side.
- [ ] Keep pulling announcements from the same place as today - this is just a visual change.
- [ ] Keep the auto-rotate working, restyled with no visible dots, arrows, or counter.
- [ ] Fix the keyboard navigation so it only responds when the banner is focused, instead of listening for arrow keys anywhere on the page - or drop keyboard navigation, since there's no longer a visible control for it to pair with.
- [ ] Remove the per-type colors and icons - one plain black bar for every announcement type - while keeping each type's existing button-link behavior working exactly as it does today.
- [ ] Decide what the button shows for an announcement that has neither a specific link nor the resource-page routing case - today it shows nothing, which shouldn't carry over to a design that always shows a button.
- [ ] Add the new "New:" label text in both Arabic and English.

## Out of Scope
- Changing how announcements are created or stored.
- The photo carousel on resource pages - a different, separate feature.
- Any other part of the homepage.

## Technical Notes
- The existing banner code is `AnnouncementsCarousel.tsx`.
- There's an existing test file for this component that checks for the controls being removed here (the prev/next buttons) - it'll need rewriting alongside the component, not just updating a few assertions.

## Acceptance Criteria
- [ ] The banner matches the new black-bar design on both desktop and mobile, with no visible dots, arrows, or counter.
- [ ] Auto-rotate still works, without visible controls, and pausing it doesn't require a mouse (e.g. it also pauses on keyboard focus, not just mouse hover).
- [ ] The announcement button still goes to the right place for every case that works today, including the one that links to a specific resource page.
- [ ] No announcement type shows a distinct color or icon - all render the same way.
- [ ] Looks and reads correctly in both Arabic and English.

Zyad-Eltayabi:
## Summary

Implemented the new announcement banner design for issue #304 while preserving the existing carousel behavior.

## Changes

1. Rebuilt the announcement banner UI to match the provided design and fit the existing page layout.
2. Added responsive behavior for different screen sizes.
3. Removed the old carousel controls and type-specific styling.
4. Preserved automatic rotation, with pause on hover and focus.
5. Kept keyboard navigation and scoped arrow-key navigation to when the banner itself is focused.
6. Ensured every announcement has a CTA, using its specific link when available and `/resources` as the fallback.
7. Preserved Arabic/English translations, RTL/LTR support, and accessibility semantics.
8. Updated the related tests to cover the new behavior and removed outdated expectations.

Closes #304 

## Summary by CodeRabbit

* **New Features**
  * Redesigned announcements as a streamlined banner experience.
  * Added automatic slide rotation with pause and resume on hover or focus.
  * Added keyboard navigation within the banner and support for announcement links to resources or specific destinations.
  * Added Arabic and English “New” labels.

* **Accessibility**
  * Added a clearly labeled carousel region for screen readers.

* **Changes**
  * Removed previous/next controls, tab indicators, and slide counters from the announcements display.


abubakr-itqan (changes requested):
Really nice rebuild overall - the black bar matches the design, auto-rotate/hover-pause/focus-pause all work, keyboard nav is properly scoped to the banner now, and the type-specific styling is gone. tsc is clean and lint doesn't add anything new.

One real gap though: `resolveHref` only checks `slide.cta_url` and falls back to `/resources` for everything else. The old component had a second case - a `breaking_change` announcement with a `resource_id` but no `cta_url` linked to `/resources/${resource_id}`, not the generic catalog. #304's AC calls this out specifically ("The announcement button still goes to the right place for every case that works today, including the one that links to a specific resource page"), and none of the new tests cover it - CodeRabbit flagged the same thing.

Could you add that case back into `resolveHref` (checking `cta_url` first, then `type === 'breaking_change' && resource_id` falling back to `/resources/${resource_id}`, then `/resources`) and add a test for it?

Also, a smaller thing from my earlier comment on the issue - the two mock announcements with "New:" baked into the title (`mock-data.ts`) still need that stripped, since the banner now always prepends its own "New:" label. Right now they'd render as "New: New: Tafsir Ibn Kathir Digital Edition" in mock/dev mode.

Zyad-Eltayabi:
Thanks for pointing this out, and sorry for the misunderstanding.

I’ve now:

* Updated `resolveHref` to preserve the `breaking_change` + `resource_id` case when no `cta_url` is provided, while keeping `cta_url` as the first priority and `/resources` as the final fallback.
* Added a test covering this missing case.
* Removed the hardcoded `New:` prefix from the two affected mock announcement titles in `mock-data.ts`, as discussed earlier.

abubakr-itqan (approved):
Re-reviewed after your update - both fixes look right. resolveHref now checks cta_url, then the breaking_change+resource_id case, then falls back to /resources, exactly matching what the old component did. New test covers the missing case directly, and the two mock titles no longer double up on "New:". tsc clean, lint matches the existing baseline, all 17 tests pass. Merging.

